PR #10, opened 2020-04-28
INSTALL: add delta and unifdef debian packages

nickdesaulniers:
Debian packages these, so they can be added to the recommended
dependency install script.

Signed-off-by: Nick Desaulniers <ndesaulniers@google.com>

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -27,7 +27,7 @@ like this:
 
 ```
 sudo apt-get install \
-  flex build-essential
+  flex build-essential unifdef delta
 ```
 
 On FreeBSD 12.1, the prerequisites can be installed like this:


marxin:
Thanks.

nickdesaulniers:
Also, @marxin I saw your [post](https://lists.llvm.org/pipermail/llvm-dev/2020-April/141188.html) via [llvm weekly](http://llvmweekly.org/issue/330).  Sorry I'm not subscribed to llvm-dev (though it seems I am to cfe-dev, weird) so I couldn't respond there.  But I wanted to say: cool! I too run `creduce` a ton, and my one point of feedback has been that it doesn't saturate my 72 core workstation.  Looking into its sources, I found my lack of knowledge of Perl to be a barrier, too.  I think I also had trouble using my llvm tree, too, IIRC.

I will absolutely give this a shot next time I need to repro, and let you know how it goes (probably via github, unless you prefer email).

One question I had; is clang_delta an external dependency? If so, how do you plan to keep your copy from diverging? Would it be better to have it be a shared git submodule, or ...?

Also, I read your masters thesis a while ago and really liked it!

marxin:
> Also, @marxin I saw your [post](https://lists.llvm.org/pipermail/llvm-dev/2020-April/141188.html) via [llvm weekly](http://llvmweekly.org/issue/330). Sorry I'm not subscribed to llvm-dev (though it seems I am to cfe-dev, weird) so I couldn't respond there. But I wanted to say: cool!

Thank you very much!!
> I too run `creduce` a ton, and my one point of feedback has been that it doesn't saturate my 72 core workstation. Looking into its sources, I found my lack of knowledge of Perl to be a barrier, too. I think I also had trouble using my llvm tree, too, IIRC.

Funny. Exactly the same reasons motivated me to finish the Moritz's Python port.

> 
> I will absolutely give this a shot next time I need to repro, and let you know how it goes (probably via github, unless you prefer email).

I prefer both, Github would be better as it's publicly available for other users.
> 
> One question I had; is clang_delta an external dependency? If so, how do you plan to keep your copy from diverging? Would it be better to have it be a shared git submodule, or ...?

Right now not. The community is quite open for making it an external dependency and I will take some work. On the other hand, I picked couple of `clang_delta` pull requests from creduce (try-catch reduction) and I added something to address csmith-project#210. As you may know, creduce upstream is not much active right now. That's another reason why I started with C-Vise.

> 
> Also, I read your masters thesis a while ago and really liked it!

What a compliment!

nickdesaulniers:
oh, hey, look! We're both on https://github.com/csmith-project/creduce/issues/160.

marxin:
Looks we've got a common interest :+1:

PR #11, opened 2020-04-28
clang_delta: fix build for ToT LLVM

nickdesaulniers:
It seems StringRef is no longer convertible to std::string, and
clang::FileManager moved since llvm-9.

Note: ToT LLVM is currently post-llvm-10 (i.e. llvm-11).

Signed-off-by: Nick Desaulniers <ndesaulniers@google.com>

### DIFF
--- a/clang_delta/CommonRenameClassRewriteVisitor.h
+++ b/clang_delta/CommonRenameClassRewriteVisitor.h
@@ -98,7 +98,7 @@ bool CommonRenameClassRewriteVisitor<T>::VisitUsingDecl(UsingDecl *D)
     return true;
 
   IdentifierInfo *IdInfo = DeclName.getAsIdentifierInfo();
-  std::string IdName = IdInfo->getName();
+  std::string IdName = IdInfo->getName().str();
   std::string Name;
   if (getNewNameByName(IdName, Name)) {
     SourceLocation LocStart = NameInfo.getBeginLoc();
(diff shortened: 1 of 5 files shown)


nickdesaulniers:
idk if you're planning on supporting older versions of LLVM, but if so, there may be some preprocessor guards we can use to support both versions.

I'm using a ToT llvm checkout, which cmake finds and uses.

nickdesaulniers:
Not sure what to make of the presubmit test failures.  I would have expected this patch to be an issue for older llvms, but the failures all seem to occur during cmake configuration?

marxin:
Thank you for pull request.

> It seems StringRef is no longer convertible to std::string, and

Yes, you are right, it's since llvm/llvm-project@777180a32b61070a10dd330b4f038bf24e916af1.
Unfortunately, your patch breaks tests with llvm-9. I think you'need to use
https://github.com/llvm/llvm-project/blob/master/llvm/include/llvm/ADT/StringRef.h#L248

Can you please rebase your patch?

marxin:
> idk if you're planning on supporting older versions of LLVM, but if so, there may be some preprocessor guards we can use to support both versions.

I hope the `::str()` is going to work with `LLVM9`, which is the older supported version right now.

marxin:
> Not sure what to make of the presubmit test failures. I would have expected this patch to be an issue for older llvms, but the failures all seem to occur during cmake configuration?

My bad, I'm gonna fix it soon.

nickdesaulniers:
> Can you please rebase your patch?

done.

marxin:
Great. I can confirm it works with LLVM9.
I'm going to merge it as soon as it survives Travis CI.

PR #22, opened 2020-05-01
cmake: Do not fail on missing pebble

mgorny:
pebble is required only at runtime, it should not prevent the package
from being built.


marxin:
Thanks!

PR #36, opened 2020-08-09
passes/ifs.py: fix infinite loop in IfPass

trofi:
Before the change IfPass was looping on the same fine:

```
$ cvise --timestamp --remove-pass IncludesPass,UnIfDefPass,CommentsPass \
    --print-diff --debug ./test_cvise.bash *.h *.cpp
2020-08-09 11:45:38,176 DEBUG granularity reduced to 2
2020-08-09 11:45:38,211 DEBUG granularity reduced to 2
2020-08-09 11:45:41,990 DEBUG granularity reduced to 2
2020-08-09 11:45:42,211 DEBUG granularity reduced to 2
2020-08-09 11:45:46,045 DEBUG granularity reduced to 2
...
```

This happens because due to two problems:
1. `state.advance()` is not a mutating operation in `ifs.py`.
   As a result increment was lost.

2. `IfPass` iteration was not able to make progress for either
   `#if 0` or `#if 1` cases.

This change fixes [1.].

### DIFF
--- a/build_and_test.sh
+++ b/build_and_test.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/env bash
 
 BUILD_TYPE=$1
 
--- a/cvise/passes/ifs.py
+++ b/cvise/passes/ifs.py
@@ -39,10 +39,12 @@ class IfPass(AbstractPass):
 
     def advance(self, test_case, state):
         if state.value == 0:
+            state = state.copy()
             state.value = 1
         else:
-            state.advance()
-            state.value = 0
+            state = state.advance()
+            if state:
+                state.value = 0
         return state
 
     def advance_on_success(self, test_case, state):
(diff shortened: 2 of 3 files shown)


marxin:
Thank you for the report.

> This happens because due to two problems:
> 
> 1. `state.advance()` is not a mutating operation in `ifs.py`.
>    As a result increment was lost.

That's definitely a mistake. Please fix the code as `BinaryState::advance` can return `None`.

> 2. `IfPass` iteration was not able to make progress for either
>    `#if 0` or `#if 1` cases.

Can you please provide a test-cases for both cases to `cvise/tests/test_ifs.py`?

Thanks

> 
> This change fixes [1.].

trofi:
Will do.

I have one more question:

```python
    def advance(self, test_case, state):
        if state.value == 0:
            state.value = 1
        else:
            state = state.advance()
            state.value = 0
        return state
```

Here `state` is mutated inplace for `0->1` transition. Should a copy be created instead as well?

marxin:
> Here `state` is mutated inplace for `0->1` transition. Should a copy be created instead as well?

Yes, it should not be modified in place. Good point.

trofi:
Added the test to expand two `#if`s in the file. It uncovered one more bug: `#if`s outside to-be-modified chunk were mangled.

marxin:
Thank you very much!

Issue #43, opened 2021-01-26
`clang_delta`: `query-instances` with `getQualifiedNameAsString`


andrewvaughanj:
Sorry if this is a "dumb" question, but what's the status of making changes to `clang_delta` in the `cvise` repo? I am not sure how closely this repo tracks `creduce`, or if `cvise` is happy to have feature extensions to `clang_delta`.

Why am I asking? Well, I am interested (in some circumstances) in running `clang_delta` to get the instance number *as well as the function name* for a given transformation.

For some inputs, I know _a priori_ a function I want to _preserve_, but I want to replace every other function with the declaration. If `query-instances` could not only give the number of instances, but also the function _name_ for each instance, that would be awesome.

If I had this, I could do something like:

```
# query all of the instances in our original file
clang_delta --query-instances=replace-function-def-with-decl test.cpp

# record the index of the function I care about; let's say it is index 3

# replace from the first to the second (leaving the third alone!)
clang_delta --transformation=replace-function-def-with-decl --counter=1 --to-counter=2 test.cpp > intermediate.cpp

# re-query to get the remaining numbers
clang_delta --query-instances=replace-function-def-with-decl intermediate.cpp

# validate that the original function should now be index 1

# replace from the second to the total number
clang_delta --transformation=replace-function-def-with-decl --counter=2 --to-counter=99 intermediate.cpp > final.cpp

# the file final.cpp now *only* has a body for the function that was at index 3 in the original file
```

I've taken a look at this, and it seems that it is achievable by using `getQualifiedNameAsString` inside of `addOneFunctionDef`.

If I was to wrap this up as a feature request, is it something that `cvise` would be happy to accept?

andrewvaughanj:
A (very) rough implementation of what I'm talking about is here:

- https://github.com/marxin/cvise/compare/master...andrewvaughanj:dump_transformation_instances?expand=1

(a number of these changes are due to `git-clang-format`; apologies if these are unwarranted)

Output looks like this:

```
avj@stoat:~/cvise> cat test.cpp
class Moo
{
    public:
        static int foo() { return 10; }
        static int bar() { return 10; }
};

void quack(void)
{
    if (Moo::foo());
}
avj@stoat:~/cvise> ./cvise/build/clang_delta/clang_delta --query-instances=replace-function-def-with-decl test.cpp
Available transformation instances: 3
Available transformation instances:
 1 Moo::foo
 2 Moo::bar
 3 quack
```

marxin:
[...]

PR #44, opened 2021-01-26
Adding a 'preserve-routine' flag to clang_delta #43

andrewvaughanj:
This PR adds a new argument (`--preserve-routine=<string>`) to `clang_delta` that allows for `replace-function-def-with-decl` to replace all routines *apart* from the one listed in `<string>`.

It also adds three test-cases to validate this new behaviour.

Signed-off-by: Andrew V. Jones <andrewvaughanj@gmail.com>


marxin:
I've just briefly looked at the `clang_delta` implementation. I would rather suggest skipping the preserved routine both in the code the counts instances, and the code which does transformation (from-to).
Note that `C-Vise` clang_delta passes are smart as so it first attempts the following transformations for N function:

```
[1, N]
[1, N/2]
[N/2, N]
[1, N/4]
```

and so on. Note that all these steps are run in parallel.

andrewvaughanj:
Hmm, I wasn't (actually) think about this being something that `cvise` would try to do automatically (and in parallel) -- more of it being an something that a user could decide to do as a very first, initial pass.

Were you think we'd integrate this into the top-level of `cvise` so someone could specify the `preserve-routine` at that level?

For my use-case, I would be interested in running `clang_delta` (manually), removing all but the body of this one routine, and *then* running `cvise` normally (which then _would_ reduce this routine!). In this workflow, I don't need `cvise` to be aware of the `preserve-routine` option.

Given this:

- [ ] Are you still okay with the feature?
- [ ] What are your thoughts on this being "oblivious" to `cvise`? If you'd like `cvise` to be aware of it, how do you think that interaction would look?

marxin:
> Hmm, I wasn't (actually) thinking about this being something that `cvise` would try to do automatically (and in parallel) -- more of it being an something that a user could decide to do as a very first, initial (potentially manual!) step.

Well, I would not do a manual step as `replace-function-def-with-decl` runs very early in the passes:
```
00:00:00 INFO INITIAL PASSES
00:00:00 INFO ===< IncludesPass >===
00:00:00 INFO ===< UnIfDefPass >===
00:00:00 INFO ===< CommentsPass >===
00:00:00 INFO ===< IfPass >===
00:00:00 INFO ===< LineMarkersPass >===
00:00:00 INFO ===< BlankPass >===
00:00:00 INFO ===< ClangBinarySearchPass::replace-function-def-with-decl >===
```

If you want to start with `ClangBinarySearchPass::replace-function-def-with-decl`, you can use `cvise --start-with-pass=ClangBinarySearchPass::replace-function-def-with-decl`.

I've added `cvise` option for that and pushed that here:
https://github.com/marxin/cvise/tree/preserve_routine

If you want to run it manually, it's fine. You will need to add `--counter=1 --to-counter=999999999 --warn-on-counter-out-of-bounds` and it will do what you want.

> 
> Were you thinking we'd integrate this into the top-level of `cvise` so someone could specify the `preserve-routine` at that level?
> 
> For my use-case, I would be interested in running `clang_delta` (manually), removing all but the body of this one routine, and _then_ running `cvise` normally (which then _would_ reduce this routine!). In this workflow, I don't need `cvise` to be aware of the `preserve-routine` option.

As mentioned, it seems to me as an extra step.
Are you fine with the updated pull request I've made in the branch?

> 
> Given this:
> 
> * [ ]  Are you still okay with the feature?
> * [ ]  What are your thoughts on this being "oblivious" to `cvise`? If you'd like `cvise` to be aware of it, how do you think that interaction would look?

andrewvaughanj:
Your changes look great to me, and I have no issues with needing to specify a "from" and "to" value to get this to work!

I think it is a good change that if you query *and* provide a `preserve` option, that you now get the correct count (vs. my change where you'd still get all the instances).

marxin:
Great, let's merge it now.
Thank you for your cooperation.

marxin:
I have one nit: would you be able to preserve a routine that can have a complex template name?

Like for:
https://github.com/gcc-mirror/gcc/blob/master/gcc/testsuite/g++.dg/gcov/pr88045.C

```
c++filt _ZSt4moveIRNSt12_Vector_baseI4TreeIJ3FooS2_EESaIS3_EE12_Vector_implEEONSt16remove_referenceIT_E4typeEOS9_
std::remove_reference<std::_Vector_base<Tree<Foo, Foo>, std::allocator<Tree<Foo, Foo> > >::_Vector_impl&>::type&& std::move<std::_Vector_base<Tree<Foo, Foo>, std::allocator<Tree<Foo, Foo> > >::_Vector_impl&>(std::_Vector_base<Tree<Foo, Foo>, std::allocator<Tree<Foo, Foo> > >::_Vector_impl&)
```

?

andrewvaughanj:
Hmm, yes, you've got a good point. While mostly this won't be an issue, I can see why it might be nice to have this functionality.

I played around, and couldn't find a good way to get either the "full" name or the mangled name from the LLVM API. If we know how to get that "full" name, then I'm totally fine to need to give that name.

Currently, for `pr88045.C`, we have these names:

```
Foo::size
Foo::Foo
Tree::Tree<Ts...>
Tree::Tree<Ts...>
size
size
main
```

which isn't great, but also isn't entirely horrible ...

marxin:
I guess there must be a way how to get a mangled name from LLVM API. Anyway, we may want to leave it for the future.
I can also imagine a regex that can cover function names that should be preserved.

andrewvaughanj:
Interestingly, LLVM even has its own in-built regex class:

- https://llvm.org/doxygen/classllvm_1_1Regex.html

I think the idea about getting the fully qualified names + using a regex sounds great. Personally, I would rather have the _unmangled_ name (despite being more verbose), as I think that will be easier for a user.

We may also want some way to use `clang_delta` to also dump all of the routine names, so a user can see the precise syntax that `clang_delta` wants.

Shall I create a GitHub issue around these and I can take a look if I get some time?

marxin:
> We may also want some way to use `clang_delta` to also dump all of the routine names, so a user can see the precise syntax that `clang_delta` wants.
> 
> Shall I create a GitHub issue around these and I can take a look if I get some time?

Yes, that should be a useful feature for the newly added `--preserve-routine`, thanks.

PR #64, opened 2021-10-15
Fix resource leaks in tests and report them via pytest

mgorny:
Fix resource leaks due to unclosed files and add `-Wall` to pytest to have it verbosely warn about any problems.


mgorny:
I suppose the CI failures are not my fault? ;-)

PR #66, opened 2021-11-18
Discourage -Werror in README.md

avikivity:
-Werror makes some passes ineffective as the introduce warnings,
and -Werror turns those into errors.


marxin:
Thanks for your pull request. However, I would rather make `-Werror` working also for Release configurations.

avikivity:
There's a misunderstanding. I'd like to discourage -Werror for the interestingness test, not for cvise itself. I like -Werror myself (which was why I was bitten by it when I left it in the interestingness test).

marxin:
Ohh! I'm really sorry :) Sure, I like your pull request.

Issue #17, opened 2020-04-29
Windows support


JVApen:
Trying to compile this on Windows (Visual Studio 2019) doesn't work.
I use the following to configure my CMake:

```
set CC="C:\Program Files\VisualStudio\Vs2019\VC\Tools\MSVC\14.24.28314\bin\Hostx64\x64\cl.exe"
set CXX="C:\Program Files\VisualStudio\Vs2019\VC\Tools\MSVC\14.24.28314\bin\Hostx64\x64\cl.exe"
set LLVM_DIR=LLVM_BUILD\lib\cmake\llvm
set CLANG_DIR=LLVM_BUILD\lib\cmake\clang
C:\Program Files\VisualStudio\Vs2019\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe -G "Visual Studio 16 2019"  -A x64 -Thost=x64  ..\cvise-master
```

`cl : command line error D8021: invalid numeric argument '/Werror' [build\clang_delta\clang_delta.vcxproj]`

After removing this flag from source/CMakeLists.txt (line 144-145), an actual build gets started.

Later we get other errors, like:
```
source\clang_delta\ExpressionDetector.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj [build\clang_delta\clang_delta.vcxproj]
cl : command line error D8021: invalid numeric argument '/Wno-unused-function' [build\clex\clex.vcxproj]
cl : command line error D8021: invalid numeric argument '/Wno-unused-function' [build\clex\strlex.vcxproj]
```
These last 2 warnings are a result of source/clex/CMakeLists.txt

I also encountered some issues with the .l files, however, this could be linked to older hacks I've added to our lexer. (I did have to add %option nounistd to the options).

With these changes, I got the C++ code compiling.
Can we get some native windows support instead?

Note: Compile times take long with a lot of warnings. (Over 14 minutes on 1 file)

marxin:
Thank you for testing build on Windows! I'm seeking for somebody who can help me with that.
I fixed the 2 errors you mentioned and I'm curious about compiler warning you see right now?
What do you mean by native windows support? A prebuilt installer?

Yes, the clang_delta compilation takes quite some time, LLVM is a huge project.

JVApen:
Hey Marnix, I'm so glad you made this fork, I've been doing these reductions by hand the last few years. Python is a bit easier to hack than perl to work around some restrictions.

[...]

PR #70, opened 2021-12-03
Allow and document building on windows

dgrunwald:
Even with the information in #17, building cvise on windows was quite difficult for me as I am unfamiliar with cmake.

This pull request documents the steps I used and fixes some build errors.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -140,6 +140,31 @@ Note that assertions are enabled by default. To disable assertions:
 cmake ... -DENABLE_TRANS_ASSERT=OFF
 ```
 
+## Building on windows
+
+Build LLVM according to https://llvm.org/docs/GettingStarted.html#getting-the-source-code-and-building-llvm
+I used these LLVM build options:
+```
+@call "%PROGRAMFILES(x86)%\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
+mkdir build
+cd build
+"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" -S ..\llvm -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD=X86 -DCMAKE_BUILD_TYPE=Release -Thost=x64
+msbuild ALL_BUILD.vcxproj /p:Configuration=Release /p:Platform=x64 /p:VisualStudioVersion=16.0 /v:m
+```
+
+Download flex from https://github.com/lexxmark/winflexbison/releases/download/v2.5.24/win_flex_bison-2.5.24.zip
+
+```
+pip install pytest psutil pebble
+@call "%PROGRAMFILES(x86)%\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
+set LLVM_DIR=D:\src\llvm-project\build\lib\cmake\llvm
+set CLANG_DIR=D:\src\llvm-project\build\lib\cmake\clang
+"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe"  -A x64 -Thost=x64  c:\src\cvise -DCMAKE_INSTALL_PREFIX=c:\tools\cvise\
+msbuild INSTALL.vcxproj /p:Configuration=Release /p:Platform=x64 /p:VisualStudioVersion=16.0 /v:m
+```
+
+Run with: `python C:\tools\cvise\bin\cvise`
+
 ## Testing
 
 You can test the project with:
--- a/clex/CMakeLists.txt
+++ b/clex/CMakeLists.txt
@@ -54,6 +54,10 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
 set_source_files_properties(clex.c PROPERTIES COMPILE_FLAGS -Wno-unused-function)
 set_source_files_properties(strlex.c PROPERTIES COMPILE_FLAGS -Wno-unused-function)
 endif()
+if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+set_source_files_properties(clex.c PROPERTIES COMPILE_FLAGS -DYY_NO_UNISTD_H)
+set_source_files_properties(strlex.c PROPERTIES COMPILE_FLAGS -DYY_NO_UNISTD_H)
+endif()
 
 ###############################################################################
 
--- a/clex/driver.c
+++ b/clex/driver.c
@@ -20,6 +20,14 @@
 
 #include "defs.h"
 
+#ifdef _MSC_VER
+// MSVC lacks this compiler builtin
+__declspec(noreturn) void __builtin_unreachable()
+{
+    abort();
+}
+#endif
+
 struct tok_t {
   char *str;
   enum tok_kind kind;
--- a/delta/CMakeLists.txt
+++ b/delta/CMakeLists.txt
@@ -29,7 +29,13 @@ add_executable(topformflat
   ${FLEX_topformflat_scanner_OUTPUTS}
   )
 
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
+    OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
 set_source_files_properties(topformflat.c PROPERTIES COMPILE_FLAGS "-Wno-unused-function -Wno-unused-parameter")
+endif()
+if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+set_source_files_properties(topformflat.c PROPERTIES COMPILE_FLAGS -DYY_NO_UNISTD_H)
+endif()
 
 ###############################################################################
 


dgrunwald:
Note that I haven't had much success using the resulting cvise.

I tried with the interestingness test:
```bat
cl.exe /c test.cpp || exit 1
```
and the simple test.cpp:
```cpp
int dummy() {
     return 1; 
}
int main() {
     return dummy(); 
}
```

Result:
```
C:\\temp\\cvise-test>python C:\tools\cvise\bin\cvise check.bat test.cpp -n 1 --debug
00:00:00 ERROR Prereqs not found for pass IfPass
00:00:00 DEBUG perform sanity check...
00:00:00 DEBUG sanity check tmpdir = C:\Users\...\AppData\Local\Temp\cvise-sanity-o_4urtoz
00:00:00 DEBUG sanity check successful
00:00:00 INFO ===< 19012 >===
00:00:00 INFO running 1 interestingness test in parallel
00:00:00 INFO INITIAL PASSES
00:00:00 DEBUG Creating pass root folder: C:\Users\dg\AppData\Local\Temp\cvise-IncludesPass-aa7yodn_
00:00:00 INFO ===< IncludesPass >===
00:00:01 DEBUG Creating pass root folder: C:\Users\dg\AppData\Local\Temp\cvise-CommentsPass-30ebektj
00:00:01 INFO ===< CommentsPass >===
00:00:01 ERROR Skipping IfPass
00:00:01 DEBUG Creating pass root folder: C:\Users\dg\AppData\Local\Temp\cvise-LineMarkersPass-xx98ovnb
00:00:01 INFO ===< LineMarkersPass >===
00:00:01 DEBUG Creating pass root folder: C:\Users\dg\AppData\Local\Temp\cvise-BlankPass-6u3zmcat
00:00:01 INFO ===< BlankPass >===
00:00:02 DEBUG Creating pass root folder: C:\Users\dg\AppData\Local\Temp\cvise-ClangBinarySearchPass-replace-function-def-with-decl-2xg5mlui
00:00:02 INFO ===< ClangBinarySearchPass::replace-function-def-with-decl >===
00:00:02 DEBUG available transformation opportunities for c++98: 2, took: 0.02 s
00:00:02 DEBUG available transformation opportunities for c++11: 2, took: 0.02 s
00:00:02 DEBUG available transformation opportunities for c++14: 2, took: 0.00 s
00:00:02 DEBUG available transformation opportunities for c++17: 2, took: 0.02 s
00:00:02 DEBUG available transformation opportunities for c++20: 2, took: 0.02 s
00:00:02 INFO using C++ standard: c++98 with 2 transformation opportunities
00:00:02 DEBUG granularity reduced to 1
00:00:02 DEBUG ***ADVANCE*** from 0 to 1 with chunk 1
00:00:02 DEBUG Creating pass root folder: C:\Users\dg\AppData\Local\Temp\cvise-ClangBinarySearchPass-remove-unused-function-yo_8s8_s
00:00:02 INFO ===< ClangBinarySearchPass::remove-unused-function >===
00:00:02 DEBUG available transformation opportunities for c++98: 0, took: 0.02 s
00:00:02 DEBUG available transformation opportunities for c++11: 0, took: 0.00 s
00:00:02 DEBUG available transformation opportunities for c++14: 0, took: 0.02 s
00:00:02 DEBUG available transformation opportunities for c++17: 0, took: 0.02 s
00:00:02 DEBUG available transformation opportunities for c++20: 0, took: 0.00 s
00:00:02 INFO using C++ standard: c++98 with 0 transformation opportunities
00:00:02 DEBUG Creating pass root folder: C:\Users\dg\AppData\Local\Temp\cvise-LinesPass-0-zp8py5q6
00:00:02 INFO ===< LinesPass::0 >===
00:00:02 DEBUG perform sanity check...
00:00:02 DEBUG sanity check tmpdir = C:\Users\dg\AppData\Local\Temp\cvise-sanity-1kon7djj
00:00:03 DEBUG sanity check successful
Traceback (most recent call last):
  File "C:\tools\cvise\bin\cvise", line 305, in <module>
    reducer.reduce(pass_group, skip_initial=args.skip_initial_passes)
  File "C:\Tools/cvise/share\cvise\cvise.py", line 143, in reduce
    self._run_additional_passes(pass_group['first'])
  File "C:\Tools/cvise/share\cvise\cvise.py", line 166, in _run_additional_passes
    self.test_manager.run_pass(p)
  File "C:\Tools/cvise/share\cvise\utils\testing.py", line 506, in run_pass
    self.state = self.current_pass.new(self.current_test_case, self.check_sanity)
  File "C:\Tools/cvise/share\cvise\passes\lines.py", line 54, in new
    self.__format(test_case, check_sanity)
  File "C:\Tools/cvise/share\cvise\passes\lines.py", line 36, in __format
    os.unlink(backup.name)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\temp\\cvise-test\\tmph4y1034u'
```

I would have expected replace-function-def-with-decl to succeed in a reduction, but somehow that didn't work.
Also there is a crash when another program like windows defender blocks the deletion of the directory.
Unfortunately I've spent enough time fighting/learning cmake that I've run out of time and won't continue from here in the near future.

marxin:
Thank you very much for the attempt and I'm willing to merge your pull request!

marxin:
> I would have expected replace-function-def-with-decl to succeed in a reduction, but somehow that didn't work.

You can invoke that manually with: `./clang_delta x.c --transformation=replace-function-def-with-decl --counter=1 --to-counter=2` and see the output.

marxin:
> Also there is a crash when another program like windows defender blocks the deletion of the directory.

Hm, are you sure it's really windows defender?

dgrunwald:
> Hm, are you sure it's really windows defender?

No I'm not sure, that was just my first guess. There's a lot of background stuff in Windows that might be holding on to directory handles.
I've seen errors like this occur (with other tools) on machines where most other services (search indexer etc.) were disabled.

I've only seen `ERROR_ACCESS_DENIED` occur with files when another process has the file open. Not sure what that could have been.
Deleting empty directories tends to non-deterministically trigger `ERROR_DIR_NOT_EMPTY` (whenever there's another process trying to concurrently list the directory) -- this one is fundamentally unavoidable on Windows and requires a retry-loop.

marxin:
Well, what C-Vise does during reduction: it spawns multiple processes with interestingness test and waits for the first one to finish. Then other processes are canceled and the top-level working directory is removed. That means there can be still running processes that access the files. Such an approach works fine on Linux but may cause troubles on Windows. (Un)fortunately, I don't use Windows system, so can't debug that ..

dgrunwald:
The problems with [file deletion itself being asynchronous](https://www.youtube.com/watch?v=uhRWMGBjlO8&t=534s) in Windows can be worked around in a number of ways.
But I guess the process cancellation is also asynchronous, and other processes (e.g. children of the interestingness test) might still have files open without specifying `FILE_SHARE_DELETE`. It's also impossible to delete a directory if there is a running process using it as current working directory.
In that case there is no solution other than waiting for those processes to quit.

I guess cvise could catch the error and then retry deleting at a later point in time (e.g. after the next reduction step).

marxin:
Looking at the backtrace:
https://github.com/marxin/cvise/blob/2aa333e3e057a1a5c097ce7d3dcd79a4dc192903/cvise/passes/lines.py#L31-L36

It's strange that one can't unlink the temporary file. After the return from `check_sanity`, the file should not be used by another process.

marxin:
> A `backup.close()` call would help (though that brings up the question: why not use `delete=True` instead?).

Good point, I've just rewritten the code in the pass. Can you please test it now if it makes a progress? If so, I can rework other places..

dgrunwald:
Weirdly enough, now the file copy fails:
```
00:00:03 INFO ===< LinesPass::0 >===
Traceback (most recent call last):
  File "c:\Tools\cvise\bin\cvise", line 305, in <module>
    reducer.reduce(pass_group, skip_initial=args.skip_initial_passes)
  File "C:\Tools/cvise/share\cvise\cvise.py", line 143, in reduce
    self._run_additional_passes(pass_group['first'])
  File "C:\Tools/cvise/share\cvise\cvise.py", line 166, in _run_additional_passes
    self.test_manager.run_pass(p)
  File "C:\Tools/cvise/share\cvise\utils\testing.py", line 506, in run_pass
    self.state = self.current_pass.new(self.current_test_case, self.check_sanity)
  File "C:\Tools/cvise/share\cvise\passes\lines.py", line 53, in new
    self.__format(test_case, check_sanity)
  File "C:\Tools/cvise/share\cvise\passes\lines.py", line 32, in __format
    shutil.copy(test_case, backup.name)
  File "C:\Python310\lib\shutil.py", line 417, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "C:\Python310\lib\shutil.py", line 256, in copyfile
    with open(dst, 'wb') as fdst:
PermissionError: [Errno 13] Permission denied: 'C:\\temp\\creduce\\tmp3ouuezhu'
```

Maybe the `delete=True` parameter somehow also influences the `FILE_SHARE_WRITE` mode?
It's a bit unclear how the Python API maps to the Win32 concepts...

marxin:
> Maybe the `delete=True` parameter somehow also influences the `FILE_SHARE_WRITE` mode? It's a bit unclear how the Python API maps to the Win32 concepts...

Hmm, that's strange. It seems the `FILE_SHARE_WRITE` mode is not easily accessible from Python :/

dgrunwald:
Note that the copy was successful previously when you were using `delete=False`.
I think the problem is unrelated to `FILE_SHARE_WRITE` (which I think Python always uses). Rather, it's that with `delete=True` the file is marked as delete-on-close, which apparently prevents any future `open()` unless that future open also allows deletion (`FILE_SHARE_DELETE`).
So using `delete=False` and instead deleting the file in a custom `try-finally` should work (as long as the file handle is closed prior to the deletion).

PR #79, opened 2022-09-08
class-template-to-class transformation was not working with out-of-line method definitions

strimo378:
The class-template-to-class transformation did not remove the template keyword from out-of-line methods.

```
template <class> class j {
public:
	void func();

	template <class c>
	void func2(c p);
};

template <class m> void j<m>::func() {
	return *this;
}
```

was translated to

```
class j {
public:
	void func();

	template <class c>
	void func2(c p);
};

template <class m> void j::func() {
	return *this;
}
```

the template specifier of `j::func` was not removed.

### DIFF
--- a/clang_delta/CMakeLists.txt
+++ b/clang_delta/CMakeLists.txt
@@ -55,6 +55,8 @@ set(SOURCE_FILES
   "/tests/class-to-struct/class-to-struct1.C"
   "/tests/class-to-struct/class-to-struct1.output"
   "/tests/class-to-struct/class-to-struct-forward.C"
+  "/tests/class-template-to-class/test1.cc"
+  "/tests/class-template-to-class/test1.output"
   "/tests/copy-propagation/copy1.cpp"
   "/tests/copy-propagation/copy1.output"
   "/tests/copy-propagation/copy2.cpp"
--- a/clang_delta/ClassTemplateToClass.cpp
+++ b/clang_delta/ClassTemplateToClass.cpp
@@ -55,6 +55,7 @@ public:
   { }
 
   bool VisitTemplateSpecializationTypeLoc(TemplateSpecializationTypeLoc Loc);
+  bool VisitCXXMethodDecl(CXXMethodDecl* MD);
 
 private:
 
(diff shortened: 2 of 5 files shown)


clang_delta/CMakeLists.txt
```diff
@@ -352,6 +354,12 @@ foreach(file IN LISTS SOURCE_FILES)
   configure_one_file(${file})
 endforeach()
 
+if (MSVC)
```
marxin, line 357: Is this one a necessary change?

marxin (changes requested):
Cool, I like your improvement! I am fine with the change except one nit.

marxin:
And please squash the changes into a single commit.

clang_delta/CMakeLists.txt
strimo378: For me it is not compiling on MSVC without \bigobj

marxin: I see, so keep this commit and squash it to 2 commits.

strimo378:
done ... I removed the /bigobj for now

I will provide more merge requests

marxin:
> I will provide more merge requests

Cool, please do so. Btw. does the tool work correctly for you on Windows?

strimo378:
Sry, cannot tell you. I use the tool on Linux. I only use Windows to develop the clang_delta tool on Visual Studio.

Another thing. Currently, I am facing the problem that sometimes my Debian creduce provides better results that cvise but I could not find out why. All the debugging options not really help me to find it out.

There is definitely some more work required to better resolve hardcore template stuff. Maybe I will provide some more transformations into that direction…

Von: Martin Liška ***@***.***>
Gesendet: Friday, 9 September 2022 14:59
An: marxin/cvise ***@***.***>
Cc: Timo Stripf ***@***.***>; Author ***@***.***>
Betreff: Re: [marxin/cvise] class-template-to-class transformation was not working with out-of-line method definitions (PR #79)


I will provide more merge requests

Cool, please do so. Btw. does the tool work correctly for you on Windows?

marxin:
> Cool, please do so. Btw. does the tool work correctly for you on Windows?

I don't know as I use Linux only. But there was some attempt in the direction, but I think the Windows support is not complete yet.

PR #80, opened 2022-09-09
return-void transformation discarded function specifiers, attributes and scope qualifiers

strimo378:
The return-void transformation replaced more than the return type. 

E.g.

`static int func()` => `void func()`
`int classname::func()` => `void func()`

In particular, omitting the class name prevented the removal of the return value for out-of-line member function definitions.


marxin:
Cool, thanks a lot for the pull request!

PR #81, opened 2022-09-10
instantiate-template-param was not working for parameters inside anonymous templates

strimo378:
See test case


marxin:
Cool, thanks a lot for the pull request!

PR #83, opened 2022-09-12
Added member-to-global pass that moves declarations within a record (class or struct) in front of the record

strimo378:
Added member-to-global pass that tries to move declaration within a record in front of the record.

A supports and was tested with typedefs/using, static/non-static functions, static/non-static variables.


marxin:
Thanks!

PR #84, opened 2022-09-14
Added rename-operator pass that converts operator functions (e.g. operator +) to regular functions

strimo378:
A pass to rename operator functions (e.g. operator +) to regular function names op1, op2, ...
Relevant operators are replaced by function calls, e.g. a + b => a.op1(b).
The pass renames all functions with --count=1 and single functions with >= 2.


marxin:
Thanks for that!

PR #85, opened 2022-09-28
CMake: Fixed compilation, enabled parallel builds and deactivated unnecessary warnings for MSVC

strimo378:
Improve compilation on MSVC

/bigobj is required for me
/mp enables parallel build ... otherwise build is REALLY slow
Deactivated warnings are copied from LLVM project


marxin:
Thanks for it!

PR #91, opened 2022-12-08
Fixed problem with infinity loop caused by rename-fun transformation

strimo378:


strimo378:
I also had compilation problems with current master caused by -Wno-error=mismatched-new-delete

`cc1plus: error: ‘-Wno-error=mismatched-new-delete’: no option ‘-Wmismatched-new-delete’`

I think you must add the option compiler specific

marxin:
> I also had compilation problems with current master caused by -Wno-error=mismatched-new-delete
> 
> `cc1plus: error: ‘-Wno-error=mismatched-new-delete’: no option ‘-Wmismatched-new-delete’`

Yeah, good point! I've just pushed a revision that fixes that.

marxin:
Well, I see it really made a renaming:

```
$ ./clang_delta/clang_delta --transformation=rename-fun --counter=1 /tmp/rename.c > /tmp/rename2.c && diff -u /tmp/rename.c /tmp/rename2.c
--- /tmp/rename.c	2022-12-08 09:45:14.231118950 +0100
+++ /tmp/rename2.c	2022-12-08 09:48:03.187305421 +0100
@@ -4,7 +4,7 @@
 template < class a > void fn1(a &&);
 template < class, class > struct b { void e(int &&); };
 template < class a, class d > void b< a, d >::e(int &&g) { fn1(g); }
-void fn3() {
+void fn1() {
   b< int > h;
   h.e(0);
 }
$ ./clang_delta/clang_delta --query-instances=rename-fun /tmp/rename2.c
Available transformation instances: 0
```

So it seems correst to me.

strimo378:
Hmm, maybe it depends on the clang version used. What happens if you remove the function that it renames?

strimo378:
`test_clang_delta.py::TestClangDelta::test_rename_fun_test2 PASSED`

is the test case failing for you?

marxin:
Then I get:

```
$ cat /tmp/rename.c
// Check that rename-fun cannot rename anything within this test case
// Maybe in future rename-fun can also rename templates functions
template < class a, class = a > class b;
template < class a > void fn1(a &&);
template < class, class > struct b { void e(int &&); };
template < class a, class d > void b< a, d >::e(int &&g) { fn1(g); }
$ ./clang_delta/clang_delta --query-instances=rename-fun /tmp/rename.c
Available transformation instances: 0

ldd ./clang_delta/clang_delta | grep LLVM
	libLLVM.so.15 => /lib64/libLLVM.so.15 (0x00007fffed200000)
```

What LLVM version do you use? Note, the usage of `isConsecutiveNumbersFromOne` is important as it distinguishes already renamed functions from the not renamed.

strimo378:
I am compiling against llvm-13 (with some changes)

marxin:
Hm, interesting. Anyway, is the transformation causing troubles as it converts the input to the very same output and the ClangPass is running again and again? Am I correct? Can you please show a reduction log where you see it?

marxin:
And do you have please a reduction output where I can see it's stuck? I can imagine comparing the output and skipping the same output as a successful transformation.

Yes, I can imagine merging the 2 passes.

strimo378:
Here it is ... 

[reduce.log](https://github.com/marxin/cvise/files/10186126/reduce.log)

strimo378:
Two small usability things ... 

--start-with-pass is not complaining if the specified pass is not existing. A message "Cannot find pass. Please use --list-passes to get a list of passes" would be nice.

--list-passes requires 1 additional dummy parameter

marxin:
> --start-with-pass is not complaining if the specified pass is not existing. A message "Cannot find pass. Please use --list-passes to get a list of passes" would be nice.

Nice point, implemented in 889b6970853c0e4c4a1d23a0c05325fc3ed54a1c.

> 
> --list-passes requires 1 additional dummy parameter

Yeah, this is more complicated as `test_cases` argument has `nargs='+'` argument. Probably not worth it.

strimo378:
It is not been trigger because the code is not executed.

`if test_env.success` is false and then it executes `elif test_env.result == PassResult.STOP`

marxin:
You said the transformation succeeds and thus I would expect  `test_env.success` would be `True`:
https://github.com/marxin/cvise/blob/96f841e316adad54bdce10cc2b045eaf4619e142/cvise/passes/clang.py#L36
?

marxin:
Ok, so with 415a2610bf03702aaf81d2f1dea32cb1d52cef60 and the following hackish patch I get:

```patch
diff --git a/cvise/passes/clang.py b/cvise/passes/clang.py
index 3e0a471..60c14ee 100644
--- a/cvise/passes/clang.py
+++ b/cvise/passes/clang.py
@@ -22,6 +22,9 @@ class ClangPass(AbstractPass):
     def transform(self, test_case, state, process_event_notifier):
         tmp = os.path.dirname(test_case)
         with tempfile.NamedTemporaryFile(mode='w', delete=False, dir=tmp) as tmp_file:
+            if self.arg == 'rename-fun':
+                return (PassResult.OK, state)
+
             args = [self.external_programs['clang_delta'], f'--transformation={self.arg}', f'--counter={state}']
             if self.user_clang_delta_std:
                 args.append(f'--std={self.user_clang_delta_std}')
```

```
$ echo 'int xxx() {return 0;}' > rename.c && ./cvise.py -c 'g++ rename.c -c && grep "return 0" rename.c' rename.c --renaming
...
00:00:57 INFO CLEANUP PASSES
00:00:57 INFO ===< ClangPass::rename-fun >===
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 WARNING ClangPass::rename-fun has encountered a non fatal bug: pass failed to modify the variant
00:00:57 INFO (0.0%, 22 bytes, 1 lines)
...
```

where it bails out after `MAX_CRASH_DIR` non fatal bugs. Can you please debug why you don't have `PassResult.OK` returned from the pass?

PR #93, opened 2022-12-13
Bugfix: instantiate-template-param transformation used always the type of the first template parameter

strimo378:
This my first of a series of PR for improving the instantiate-template-param pass.

It fixes a bug that always the type of the first parameter was used for replacement. In the example g was replaced by int instead of double.


marxin:
Thanks!

PR #95, opened 2022-12-14
Pass instantiate-template-param learned to remove template arguments and parameters

strimo378:
The pass instantiate-template-param can now remove single template parameters and arguments. Before it could only remove it if there was only one template parameters. Template argument removal is new.

The pass furthermore supports now handling of function declarations (without definition).


strimo378:
Btw. have you thought about running all clang passes in a loop such that all clang passes gets repeated until no modification was applied. After ClangPass::member-to-global there could be check if a previous clang pass changes something. If yes, it is executing ClangPass::remove-namespace again.

Furthermore, I think I can provide you a clang_delta version that allows to query instances of multiple passes. That could be used to speed up execution of all Clang passes by just querying the instances of all clang passes at the beginning.

marxin:
Thanks, a nice improvement.

strimo378:
> Thanks, a nice improvement.

Thx

PR #97, opened 2022-12-15
The move-function-body transformation learned to move class methods, templated functions and methods within templated classes

strimo378:
See test cases


marxin:
Very nice work, again!

Please fix build error seen with older LLVM 9:

```
/__w/cvise/cvise/clang_delta/MoveFunctionBody.cpp:105:68: error: 'class clang::CXXRecordDecl' has no member named 'getDescribedTemplateParams'; did you mean 'getDescribedTemplate'?
  105 |       if (const TemplateParameterList* ClassTPL = MD->getParent()->getDescribedTemplateParams()) {
      |                                                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

Is it something we can build conditionally with `#if LLVM_VERSION_MAJOR` ?

strimo378:
That is a convenient function. I copied it form clang source...

PR #121, opened 2023-11-15
cvise.py: call shell vial `/usr/bin/env` indirection instead of `/bin`

trofi:
My system (`NixOS`) has one well-defined shell: `/bin/sh`.

But normally programs use either `/run/current-system/sw/bin/bash` or the `bash` from `PATH`. They use `/usr/bin/env` for that.

The change calls shell via `/usr/bin/env` to use `PATH` lookup.


marxin:
Thanks for the contribution.

PR #124, opened 2024-01-29
CMakeLists changes to fix macOS build:

comex:
- Check whether `CMAKE_CXX_COMPILER_ID` contains the string `Clang` rather than being exactly equal.  This way it also matches `AppleClang`, which is what CMake uses for the default compiler on macOS; see also:

  https://cmake.org/cmake/help/latest/policy/CMP0025.html

- Add `-Wno-sign-compare` to the special options for Flex-generated .c files, because Apple's Flex generates code that produces this warning by default.  Compare to e.g.:

  https://github.com/vlm/asn1c/issues/321

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,7 +168,7 @@ check_cxx_compiler_flag(
 )
 
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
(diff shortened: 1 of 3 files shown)


marxin:
Thanks for the pull request!

PR #129, opened 2024-04-01
Fix working of --log-file parameter

dufferzafar:
We're removing all handlers after creating the basicConfig object & then only appending the syslog!

BTW, since we have a custom formatter shouldn't it be set on the file stream handler as well?

### DIFF
--- a/cvise.py
+++ b/cvise.py
@@ -361,7 +361,6 @@ if __name__ == '__main__':
     syslog = logging.StreamHandler()
     formatter = DeltaTimeFormatter(log_format)
     syslog.setFormatter(formatter)
-    logging.getLogger().handlers = []
     logging.getLogger().addHandler(syslog)
 
     pass_options = set()


marxin:
Well, if I remove the suggested line, then I would get:
```
INFO:root:===< ClangBinarySearchPass::remove-unused-function >===
00:00:00 INFO ===< ClangBinarySearchPass::remove-unused-function >===
INFO:root:using C++ standard: c++2b with 0 transformation opportunities
00:00:00 INFO using C++ standard: c++2b with 0 transformation opportunities
INFO:root:===< LinesPass::0 >===
00:00:00 INFO ===< LinesPass::0 >===
...
```

that's undesirable I guess.

dufferzafar:
Oh. I see what we're doing there. We're first using `basicConfig` which will add it's own handler & then we're also adding our own streamhandler with custom formatter.

I think we should just ditch `basicConfig` & replace it with our own calls to `setLevel` - that is all that remains!

This is what basicConfig does: https://github.com/python/cpython/blob/9dae05ee59eeba0e67af2a46f2a2907c9f8d7e4a/Lib/logging/__init__.py#L2081-L2100

marxin:
Sure, can you please help me with that? Can you prepare a PR now?

PR #131, opened 2024-04-03
Removed basicConfig & fixed working of --log-file arg

mzfr:
I followed the discussion in this(https://github.com/marxin/cvise/pull/129) thread and decided to send a PR.

With these changes, logs are both displayed on stdout and stored in the logfile when the `--log-file` parameter is provided.

Would it be more appropriate to solely store them in a file without displaying them on stdout if the parameter is provided?

### DIFF
--- a/cvise.py
+++ b/cvise.py
@@ -354,8 +354,7 @@ if __name__ == '__main__':
     else:
         log_config['level'] = getattr(logging, args.log_level.upper())
 
-    logging.getLogger().setLevel(log_config["level"])
-
+    logging.getLogger().setLevel(log_config['level'])
 
     syslog = logging.StreamHandler()
     formatter = DeltaTimeFormatter(log_format)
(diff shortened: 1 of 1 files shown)


marxin:
Thanks! Looks good to me. Please fix the `ruff` issue and I'm going to merge it:
`cvise.py:360:5: E303 [*] Too many blank lines (2)`

mzfr:
@marxin What are your thoughts about the following:

> Would it be more appropriate to solely store them in a file without displaying them on stdout if the parameter is provided?

marxin:
> Would it be more appropriate to solely store them in a file without displaying them on stdout if the parameter is provided?

Yeah, I would welcome it, it makes fully sense and it's even documented in that way.

mzfr:
The latest commit in this pull request should address the behavior we discussed.

marxin:
Note I also redirected pass statistics and the final test-case to the log file: https://github.com/marxin/cvise/commit/30a38f4991bf32e8d2baf57b9aa20025dedcd708.
Hope it's the expected behavior for users, what do you think?

Issue #110, opened 2023-04-17
Allow folders/directories in test cases


dmcdougall:
This issue isn't really an issue in the true sense of the word.  I'm just opening an issue because I wasn't sure where else to get some insight.

I noticed that passing test case locations that aren't in the current directory, and therefore contain a directory, are [not supported](https://github.com/marxin/cvise/blob/master/cvise/utils/testing.py#L159-L160).  Is there a technical reason for this limitation?  I use [multidelta](https://github.com/mpflanzer/delta) quite a bit.  My workflows usually involve minimising complex applications which often come with source trees that aren't necessarily "flat", and multidelta is happy to accept testcases with folders in the path.

I'm happy to poke around in the cvise source code and contribute a patch that would allow users to pass test cases that contain directories, but I just wanted to make sure that there wasn't some historical context behind the limitation first.  Perhaps someone has already tried this and it was too awkward to support?  Any insight would be valuable.

marxin:
All right, so I made a prototype in the following branch:
https://github.com/marxin/cvise/tree/add--copy-extra-folder

It adds a new argument `--copy-extra-folder` that copies content (recursively) of a directory to a temporary place where inter. test is run. Note the reduced test-cases may be contained in the folder and it can be also `.` folder if you want.
Please test it and give me a feedback if it's feasible or not.

marxin:
Oh, you expected only the supporting folder structure. So can you please reduce my patch candidate, probably rename the option, test it, and make a pull request that would fill your needs? Thanks.

dmcdougall:
Sure, I'd be happy to.

dmcdougall:
My attempt is in #132.

Constructive criticism is very welcome.

dmcdougall:
Fixed in #133

PR #132, opened 2024-04-13
Make CVise understand files in a filesystem tree

dmcdougall:
CVise now understands test cases that live in subdirectories in a filesystem tree.

The behaviour is to simply create the directories needed in tmpfs. And interestingness checks are already expected to handle paths to test cases.

For example, a user can now say:

`cvise <args> ./interesting.sh dir1/file1 file2 dir2/file1 dir3/file3`

This feature ought to make it easier to minimise more complicated situations involving larger applications that create complex out-of-source build trees where:

1. there are several files in different directories; and
2. files that have the same name but live in different directories.

There is some extra historical context behind this pull request.  I had initially made an inquiry about copying extra folders in #110.  In that issue, however, CVise didn't really understand that the directory is really part of the test case.  What I'm trying to do here is give the user the opportunity to minimise test cases without having to extract them all from an existing and potentially complex directory hierarchy.

I'm very open to criticism on this contribution, so please feel free to tear it apart.  Or if this contribution isn't something that the community is interested in, I'm happy for it to be rejected.

### DIFF
--- a/cvise/utils/testing.py
+++ b/cvise/utils/testing.py
@@ -48,10 +48,12 @@ class TestEnvironment:
         folder,
         test_case,
         additional_files,
+        test_case_dirs,
         transform,
         pid_queue=None,
     ):
         self.test_case = None
+        self.test_case_dirs = dict()
         self.additional_files = set()
         self.state = state
         self.folder = folder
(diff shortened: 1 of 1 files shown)


marxin:
Thanks for the pull request. I decided to use this opportunity and make a bigger refactoring related to `pathlib` in the `testing.py` file. Having that, I added the requested functionality. Please test my #133, does it work for you?

dmcdougall:
> I added the requested functionality

Thank you.

> Please test my https://github.com/marxin/cvise/pull/133

Will do.  Thanks again.

PR #134, opened 2024-04-15
Implement a stopping threshold

dmcdougall:
Sometimes it's enough to reduce a testcase by, say, 95%.  This commit lets the user specify this threshold and CVise doesn't need to spend time trying to make progress until it can no longer do so.

For example:

`cvise <args> --stopping-threshold 0.95 ./interesting.sh test_case`

This can speed up CVise in some cases.

Constructive feedback is very welcome.


marxin:
I like the idea! Please run `ruff format`.

marxin:
Lemme fix the format.

PR #158, opened 2025-01-15
Delete assert of exitcode in TestManager

### LINKED ISSUE #157: assert test_env.exitcode
I observed hitting `assert test_env.exitcode` in `TestManager.process_done_futures`: https://github.com/marxin/cvise/blob/29fc072fdba3abfe1c1d763af3ae84509276bef0/cvise/utils/testing.py#L423

emaxx-google:
The exitcode isn't guaranteed to be set whenever |result| is |OK|, because the code that sets it (the caller of run_test()) might be skipped on exception.

The code after the assert seems to correctly handle the case when |exitcode| is |None|, so this commit just deletes the assert. This fixes #157.

### DIFF
--- a/cvise/utils/testing.py
+++ b/cvise/utils/testing.py
@@ -420,7 +420,6 @@ class TestManager:
                 else:
                     self.pass_statistic.add_failure(self.current_pass)
                     if test_env.result == PassResult.OK:
-                        assert test_env.exitcode
                         if self.also_interesting is not None and test_env.exitcode == self.also_interesting:
                             self.save_extra_dir(test_env.test_case_path)
                     elif test_env.result == PassResult.STOP:


intrigus-lgtm:
Why not set `self.result` to `PassResult.INVALID` when `self.exitcode` is `None` here?
https://github.com/marxin/cvise/blob/29fc072fdba3abfe1c1d763af3ae84509276bef0/cvise/utils/testing.py#L106

Shouldn't that also work?
This way we could keep the assert.

emaxx-google:
> Why not set `self.result` to `PassResult.INVALID` when `self.exitcode` is `None` here?
> 
> https://github.com/marxin/cvise/blob/29fc072fdba3abfe1c1d763af3ae84509276bef0/cvise/utils/testing.py#L106
> 
> Shouldn't that also work? This way we could keep the assert.

Technically this would work too, yes. The small downside is that we'd overwrite the value actually returned from the pass - the exception in `run_test()` seems to be an artifact of task cancellation and is really not the "fault" of the pass. But as far I can see this slight semantic change wouldn't matter for the existing code.

I'm happy to change the PR to undo the assert deletion if the consensus is this alternative solution is better.

marxin:
I am fine with the assert removal, thank you for the PR.

Issue #163, opened 2025-01-21
Trimming spaces after topformflat?


emaxx-google:
The `topformflat` tool used in the `lines` pass can either add new line breaks lines or replace some of them with spaces. This often results in a noticeable increase of the file size - I observed around +1% in typical real-world reproducers.

While it's not a big problem on its own, the extraneous whitespaces do cause a bit of overhead, especially when one runs with a config that has multiple series of `lines` passes (e.g., with `max-transforms`).

Would it be reasonable to, for instance, trim leading&trailing whitespaces from the output of `topformflat`? I think this can be added in a safe way - we can use `check_sanity` to validate that the trimming didn't break the reproducer, and fall back to the untrimmed output if it did.

marxin:
Sure, go for it, it fully makes sense. One aspect of keeping the newlines is that one can see a reasonable diff when running with `--print-diff `. What do you think?

PR #164, opened 2025-01-22
Add tests for LinesPass

emaxx-google:
Add basic unit tests.

This is preparation for testing the fix for #163.


marxin:
Thanks for the changes. Just a nit: for the future, we might want to get rid of `os.path.join` and rather use `pathlib`.

emaxx-google:
> Thanks for the changes. Just a nit: for the future, we might want to get rid of `os.path.join` and rather use `pathlib`.

acknowledged, thanks

PR #165, opened 2025-01-22
Trim leading/trailing spaces after topformflat

### LINKED ISSUE #163: Trimming spaces after topformflat?

emaxx-google:
This reduces the overhead slightly and prevents the files from
growing noticeably when the lines passes start. This fixes #163.

Note that this new logic has a safeguard that falls back to the
previous behavior (which takes the topformflat's output as-is) in
case the interestingness script starts failing.

### DIFF
--- a/cvise/passes/lines.py
+++ b/cvise/passes/lines.py
@@ -19,31 +19,40 @@ class LinesPass(AbstractPass):
         with (
             CloseableTemporaryFile(mode='w+', dir=tmp) as backup,
             CloseableTemporaryFile(mode='w+', dir=tmp) as tmp_file,
+            CloseableTemporaryFile(mode='w+', dir=tmp) as stripped_tmp_file,
         ):
             backup.close()
             with open(test_case) as in_file:
                 try:
                     cmd = [self.external_programs['topformflat'], self.arg]
-                    proc = subprocess.run(cmd, stdin=in_file, capture_output=True, text=True)
+                    with subprocess.Popen(cmd, stdin=in_file, stdout=subprocess.PIPE, text=True) as proc:
+                        for line in proc.stdout:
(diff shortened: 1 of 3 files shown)


cvise/passes/lines.py
```diff
@@ -19,31 +19,39 @@ def __format(self, test_case, check_sanity):
...
             CloseableTemporaryFile(mode='w+', dir=tmp) as backup,
             CloseableTemporaryFile(mode='w+', dir=tmp) as tmp_file,
+            CloseableTemporaryFile(mode='w+', dir=tmp) as stripped_tmp_file,
         ):
             backup.close()
             with open(test_case) as in_file:
                 try:
                     cmd = [self.external_programs['topformflat'], self.arg]
-                    proc = subprocess.run(cmd, stdin=in_file, capture_output=True, text=True)
+                    with subprocess.Popen(cmd, stdin=in_file, stdout=subprocess.PIPE, text=True) as proc:
```
emaxx-google, line 28: I took the liberty to change this to read stdout per line, instead of accumulating everything in the memory. It's not a big deal, but the optimization felt easily doable...

Issue #169, opened 2025-03-27
Proposal: hint-based highly parallelizable, pluggable reduction


emaxx-google:
Proposal: hint-based highly parallelizable pluggable reduction

After experimenting with C-Vise and C-Reduce, I noticed the following aspects impede the reduction performance:

* Only one reduction heuristic is attempted at a time:
  + this makes the performance very sensitive to the choice and ordering of heuristics;
* A single heuristic keeps being used even when it begins finding only very small reductions:
  + while the user can intervene using a hotkey, this requires monitoring from an experienced human;
* When multiple reductions are discovered concurrently, only one is picked up:
  + this implies the speed is bounded by the size of a single reduction, despite that many different reductions are often discovered quickly and they could typically be applied together safely;
* Some of the important heuristics rely on reformatting the input file:
  + as this isn't 100% reliable, this requires spending additional time on the "sanity check", and in cases the reformatted input fails the heuristic's performance suffers severely;
* Integrating new heuristics is nontrivial:
  + tight coupling with the codebase, the necessity to repeat common logic (binary search), in addition to the ordering/parallelism issues mentioned above, make it almost impractical despite potential speedups for particular use cases (new input language, lightweight grammar parsers, multi-file inputs, reducing command-line flags, etc.).

---

I'd like to propose the following changes in order to address these issues:

1. Change each heuristic to **only generate descriptions of reduction possibilities ("hints")** instead of directly modifying the input file.
   + A hint may look like "delete bytes with indices X..Y", or "replace bytes with indices X..Y with text T", or a list of these, etc.
   + Example for the "RemoveUnusedFunction" heuristic:
     ```c++
     // vvv  hint1  vvv
     void f() {
       int x = 42;
     }
     // ^^^  hint1  ^^^

     class A {
       // vvv  hint2 sub-item1  vvv
       void g();
       // ^^^  hint2 sub-item1  ^^^
     };

     // vvv  hint2 sub-item2  vvv
     void A::g() {
     }
     // ^^^  hint2 sub-item2  ^^^
     ```
   + The actual protocol is TBD, but, for example, it could be JSON lines with items like `[{"l": 42, "r": 84}]`.

2. Allow the user to specify **additional heuristic generation tools**.
   + Such a tool could be developed separately and using any language - all it needs is to adhere to the output format described above.

3. Implement a **generic logic that attempts reductions by picking subsets of hints**.
   + Applying hints is a pure text editing task (abstracted away from the the reduction heuristic or language).
[...]

PR #175, opened 2025-06-20
hints: Introduce basic hint (text deletion) format & helpers

emaxx-google:
Hints are the proposed format for being used by reduction heuristics. A hint describes one or several modifications (e.g., deletion of text at particular location) to be attempted on the test case during reduction.

This uniform intermediate representation will bring the following benefits, both for performance as well as versatility:

* It's easier to implement reduction heuristics - they don't need to take care of editing the text, implementing binary search for the sake of performance, etc.

* Allow different heuristics to be attempted concurrently, in an interleaving fashion.

* Successful reductions that were concurrently discovered can be folded (merged) instead of proceeding with only one reduction.

This commit lays the groundwork for #169 - the format (JSON Schemas) and implementation of the most primitive hint format: simple text deletion.


.github/workflows/build-llvm-nightly.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     steps:
       - run: apt-get update
-      - run: apt-get -qq install -y gcc g++ wget lsb-release wget software-properties-common gnupg git cmake flex python3-pebble python3-psutil python3-chardet python3-pytest vim unifdef sudo
+      - run: apt-get -qq install -y gcc g++ wget lsb-release wget software-properties-common gnupg git cmake flex python3-pebble python3-psutil python3-chardet python3-pytest python3-jsonschema vim unifdef sudo
```
VitaNuo, line 24: Not sure if that change is still relevant, but at some point `python3-pyzstd` was also necessary.

cvise/utils/hint.py
```diff
@@ -0,0 +1,101 @@
...
+    for p in merged_places:
+        left = p['l']
+        right = p['r']
+        assert start_pos <= left < len(orig_data)
+        assert left < right <= len(orig_data)
+        # Add the chunk up to the current beginning unmodified.
+        new_data += orig_data[start_pos:left]
+        # Delete the chunk inside the current edit.
+        start_pos = right
+    # Add unmodified the chunk after the last edit"s end.
```
VitaNuo, line 68: Nit: `add the unmodified chunk after the last edit end`.

cvise/utils/hint.py
```diff
@@ -0,0 +1,101 @@
...
+    new_data += orig_data[start_pos:]
+
+    with open(file, 'w') as f:
+        f.write(new_data)
+
+
+def merge_overlapping_places(places: Sequence[object]) -> Sequence[object]:
+    """Returns non-overlapping hint places, merging places where necessary."""
```
VitaNuo, line 76: What's the difference between a "hint" and a "hint place"? It might be the same.

cvise/utils/hint.py
```diff
@@ -0,0 +1,101 @@
...
+        },
+    },
+    'required': ['l', 'r'],
+}
+
+HINT_SCHEMA = {
+    'description': 'Hint object - a description of modification(s) of the input',
+    'type': 'object',
+    'properties': {
+        'p': {
```
VitaNuo, line 38: Why not just "h" for "hint"? I am a little confused by the "place" concept.

cvise/utils/hint.py
emaxx-google: What I'm trying to model here is that a single logical change that consists of multiple modifications in different places. E.g., deletion of a function in the `RemoveUnusedFunction` heuristic that removes not only the function definition but also all declaration(s).

cvise/utils/hint.py
emaxx-google: Maybe the term "patches" would be more intuitive? A single hint that consists of one or multiple patches (to be all applied together).

cvise/utils/hint.py
VitaNuo: Ok got it. I guess `patch` is a better term indeed.

.github/workflows/build-llvm-nightly.yml
emaxx-google: Ack - It'll be part of subsequent PRs that implement the on-disk storage of hints.

emaxx-google:
Thanks!

PR #176, opened 2025-06-20
ruff: Assume Python 3.8

emaxx-google:
This is to prevent the Ruff linter from complaining on valid code needed for
compatibility with Python 3.8 - e.g., the usage of types imported from the
"typing" module.

Ruff assumes Python 3.9 by default, meanwhile C-Vise according to is
INSTALL.md only requires 3.8.

This commit is preparation for #175.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,5 +64,5 @@ jobs:
     - run: zypper -n install python3-pip
     - uses: actions/checkout@v4
     - run: pip install --break-system-packages ruff
-    - run: ruff check
+    - run: ruff check --target-version py38
     - run: ruff format --diff


VitaNuo:
Is there a reason you've decided to fix this instead of updating the list of C-Vise prerequisites?
AFAIU the statement that it can run under Python 3.8 seems incorrect.

emaxx-google:
> Is there a reason you've decided to fix this instead of updating the list of C-Vise prerequisites? AFAIU the statement that it can run under Python 3.8 seems incorrect.

It could remain unnoticed because affected people might not use the latest version of C-Vise - the incompatible change #154 landed end of 2024. I'm not sure about the impact (e.g., IIUC Ubuntu 20.04 LTS used Python 3.8 and is still within extended support), but I'm a bit reluctant to bump system requirements without a hard reason. The fix is quite simple too.

VitaNuo:
> The fix is quite simple too.

Ok, SG as long as this PR and https://github.com/marxin/cvise/pull/177 fix it.

VitaNuo (approved):

PR #180, opened 2025-06-23
Refactor arg passing for Pass.new()

emaxx-google:
Change all subclasses to use **kwargs instead of mentioning the unused parameter (using the real name or using "_").

This refactoring allows us to easily add/remove parameters needed for specific subclasses without having to change unrelated classes. This is preparation work for adding new logic for #169.

### DIFF
--- a/cvise/passes/lines.py
+++ b/cvise/passes/lines.py
@@ -61,7 +61,7 @@ class LinesPass(AbstractPass):
             lines = in_file.readlines()
             return len(lines)
 
-    def new(self, test_case, check_sanity=None):
+    def new(self, test_case, check_sanity=None, **kwargs):
         self.bailout = False
         # None means no topformflat
         if self.arg != 'None':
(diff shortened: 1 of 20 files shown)


cvise/passes/lines.py
```diff
@@ -61,7 +61,7 @@ def __count_instances(self, test_case):
             lines = in_file.readlines()
             return len(lines)
 
-    def new(self, test_case, check_sanity=None):
+    def new(self, test_case, check_sanity=None, **kwargs):
```
VitaNuo, line 64: Why did you keep `check_sanity` here and remove it in other cases?

VitaNuo (changes requested):

VitaNuo (approved):

cvise/passes/lines.py
emaxx-google: Yes the main goal was to let unnecessary parameters end up in this catch-all `kwargs`, so that code is less cluttered and future PRs removing `check_sanity` or adding new parameters are simpler.

VitaNuo: > Yes the main goal was to let unnecessary parameters end up in this catch-all kwargs

Ah cool, I assumed you wanted necessary parameters to land there as well. Got it, thanks.

PR #182, opened 2025-06-23
hints: Skeleton class for hint-based pass

emaxx-google:
Implement a base class with common logic for hint-based passes.

The subclasses are intended to only implement the hint generation according
to their heuristic. The generic class wires this up with the usual interface of a
pass, taking care of storing hints in a temp file, iterating through them using
a binary-search strategy, and regenerating hints after a successful reduction.

This is part of the effort tracked by #169.

### DIFF
--- a/cvise/passes/hint_based.py
+++ b/cvise/passes/hint_based.py
@@ -35,7 +35,7 @@ class HintState:
 
 
 class HintBasedPass(AbstractPass, metaclass=ABCMeta):
-    """Based class for hint-based passes.
+    """Base class for hint-based passes.
 
     Subclasses must implement the generate_hints() method; the
     new/transform/advance operations are taken care by the generic logic here.
@@ -45,16 +45,18 @@ class HintBasedPass(AbstractPass, metaclass=ABCMeta):
     def generate_hints(self, test_case: Path) -> Sequence[object]:
         pass
 
-    def new(self, test_case, temp_dir, **kwargs):
+    def new(self, test_case, tmp_dir, **kwargs):
         hints = self.generate_hints(test_case)
         if not hints:
             return None
-        hints_file_path = temp_dir / HINTS_FILE_NAME
+        hints_file_path = tmp_dir / HINTS_FILE_NAME
         store_hints(hints, hints_file_path)
         return HintState.create(len(hints), hints_file_path)
 
     def transform(self, test_case, state, process_event_notifier):
-        hints = load_hints(state.hints_file_path, state.binary_state.index, state.binary_state.end())
+        hints_range_begin = state.binary_state.index
+        hints_range_end = state.binary_state.end()
+        hints = load_hints(state.hints_file_path, hints_range_begin, hints_range_end)
         apply_hints(hints, Path(test_case))
         return (PassResult.OK, state)
 
--- a/cvise/tests/test_hint_based.py
+++ b/cvise/tests/test_hint_based.py
@@ -10,6 +10,9 @@ from cvise.utils.hint import HINT_SCHEMA
 class StubHintBasedPass(HintBasedPass):
     def __init__(self, contents_to_hints: Dict[str, Sequence[object]]):
         super().__init__()
+        for hints in contents_to_hints.values():
+            for hint in hints:
+                jsonschema.validate(hint, schema=HINT_SCHEMA)
         self.contents_to_hints = contents_to_hints
 
     def generate_hints(self, test_case: Path) -> Sequence[object]:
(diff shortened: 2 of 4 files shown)


cvise/passes/hint_based.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Subclasses must implement the generate_hints() method; the
+    new/transform/advance operations are taken care by the generic logic here.
+    """
+
+    @abstractmethod
+    def generate_hints(self, test_case: Path) -> Sequence[object]:
+        pass
+
+    def new(self, test_case, temp_dir, **kwargs):
```
VitaNuo, line 48: My non-existent Python knowledge is a bit strained here.. you seem to inherit from `AbstractPass`, but I don't see how this would call the parent class constructor? And what happens to `check_sanity` parameter that seems required in the parent class constructor and doesn't have a default value?

VitaNuo, line 48: Nit: sometimes it's temp and sometimes tmp.. maybe settle on tmp?

cvise/passes/hint_based.py
```diff
@@ -0,0 +1,69 @@
...
+        hints = self.generate_hints(test_case)
+        if not hints:
+            return None
+        hints_file_path = temp_dir / HINTS_FILE_NAME
+        store_hints(hints, hints_file_path)
+        return HintState.create(len(hints), hints_file_path)
+
+    def transform(self, test_case, state, process_event_notifier):
+        hints = load_hints(state.hints_file_path, state.binary_state.index, state.binary_state.end())
```
VitaNuo, line 57: Nit: consider a version of `load_hints` that would load from start index till the end of file.

cvise/tests/test_hint_based.py
```diff
@@ -0,0 +1,108 @@
...
+        super().__init__()
+        self.contents_to_hints = contents_to_hints
+
+    def generate_hints(self, test_case: Path) -> Sequence[object]:
+        contents = test_case.read_text()
+        return self.contents_to_hints.get(contents, [])
+
+
+def test_hint_based_first_char_once(tmp_path: Path):
+    """Test the case of a sole hint."""
```
VitaNuo, line 21: Nit: `single hint`.

cvise/tests/test_hint_based.py
```diff
@@ -0,0 +1,108 @@
...
+    hint = {'p': [{'l': 0, 'r': 1}]}
+    jsonschema.validate(hint, schema=HINT_SCHEMA)
```
VitaNuo, line 23: Consider validating hints inside the pass.

cvise/tests/test_hint_based.py
```diff
@@ -0,0 +1,108 @@
...
+    pass_ = StubHintBasedPass(
+        {
+            'foo': [hint],
+            'oo': [],
```
VitaNuo, line 27: Does this line add anything useful?

cvise/tests/test_hint_based.py
```diff
@@ -0,0 +1,108 @@
...
+def test_hint_based_last_char_repeatedly(tmp_path: Path):
+    """Test the case of applying a single hint that's different every time."""
+    hint0 = {'p': [{'l': 0, 'r': 1}]}
+    hint1 = {'p': [{'l': 1, 'r': 2}]}
+    hint2 = {'p': [{'l': 2, 'r': 3}]}
+    for h in hint0, hint1, hint2:
+        jsonschema.validate(h, schema=HINT_SCHEMA)
+    pass_ = StubHintBasedPass(
+        {
+            'foo': [hint2],
```
VitaNuo, line 47: Nit: I would reverse hint indexing for better readability.

cvise/tests/test_hint_based.py
```diff
@@ -0,0 +1,108 @@
...
+    iterate_pass(pass_, test_case, temp_dir=tmp_path)
+
+    assert test_case.read_text() == ''
+
+
+def test_hint_based_all_chars_grouped(tmp_path: Path):
+    """Test the case of multiple hints to be picked up together as a group.
+
+    Currently, the grouping behavior is implemented as a binary search."""
+    hint1 = {'p': [{'l': 0, 'r': 1}]}
```
VitaNuo, line 64: Nit: in previous test case hint indexing starts from 0.

cvise/tests/test_hint_based.py
```diff
@@ -0,0 +1,108 @@
...
+    hint2 = {'p': [{'l': 1, 'r': 2}]}
+    hint3 = {'p': [{'l': 2, 'r': 3}]}
+    for h in hint1, hint2, hint3:
```
VitaNuo, line 67: Nit: extract `hint1, hint2, hint3` into a variable.

cvise/tests/test_hint_based.py
```diff
@@ -0,0 +1,108 @@
...
+
+    iterate_pass(pass_, test_case, temp_dir=tmp_path)
+
+    assert test_case.read_text() == ''
+
+
+def test_hint_based_all_chars_grouped(tmp_path: Path):
+    """Test the case of multiple hints to be picked up together as a group.
+
+    Currently, the grouping behavior is implemented as a binary search."""
```
VitaNuo, line 63: What does binary search mean here? AFAICT all hints are applicable.

cvise/tests/test_hint_based.py
```diff
@@ -0,0 +1,108 @@
...
+def test_hint_based_state_iteration(tmp_path: Path):
+    """Test advancing through multiple hints.
+
+    Unlike iterate_pass-based tests which pretend that any transformation leads
+    to a successful interestingness test and proceed immediately, here we
+    verify how different hints are attempted."""
+    hint0 = {'p': [{'l': 0, 'r': 1}]}
+    hint1 = {'p': [{'l': 1, 'r': 2}]}
+    hint3 = {'p': [{'l': 3, 'r': 4}]}
+    hint4 = {'p': [{'l': 4, 'r': 5}]}
```
VitaNuo, line 91: These hints never intersect. Could we add a hint that intersects with another / is a sub- or superset?

cvise/tests/testabstract.py
```diff
@@ -1,11 +1,24 @@
...
+
+def collect_all_transforms(pass_: AbstractPass, state, input_path: Path):
+    all_outputs = set()
+    backup = input_path.read_text()
+    while state is not None:
+        pass_.transform(input_path, state, process_event_notifier=None)
+        all_outputs.add(input_path.read_text())
+        input_path.write_text(backup)
+        state = pass_.advance(input_path, state)
+    return all_outputs
```
VitaNuo, line 24: Probably missing something, but AFAICT `transform` calls `apply_hints` which applies all hints in order. Where does it store intermediate steps?

VitaNuo (commented):
Thanks! Left a few comments (mostly around me not quite understanding how Python and the `BinaryState` thingy work ;-), will continue tomorrow.

emaxx-google:
thanks for the thorough review! addressed and/or replied.

cvise/tests/test_hint_based.py
VitaNuo: Makes sense, thanks.

cvise/tests/testabstract.py
VitaNuo: Makes sense!

PR #186, opened 2025-06-24
hints: Migrate line_markers pass

emaxx-google:
Reimplement the line_markers pass using the hints architecture -
this acts as the simple first example of a heuristic implemented that
way. This migration is intended to have no functional changes.

This is part of the effort tracked by #169.

### DIFF
--- a/cvise/passes/line_markers.py
+++ b/cvise/passes/line_markers.py
@@ -1,46 +1,21 @@
-import os
 import re
-import shutil
-import tempfile
 
-from cvise.passes.abstract import AbstractPass, BinaryState, PassResult
+from cvise.passes.hint_based import HintBasedPass
 
 
-class LineMarkersPass(AbstractPass):
+class LineMarkersPass(HintBasedPass):
     line_regex = re.compile('^\\s*#\\s*[0-9]+')
 
     def check_prerequisites(self):
         return True
 
-    def __count_instances(self, test_case):
-        count = 0
+    def generate_hints(self, test_case):
+        hints = []
         with open(test_case) as in_file:
+            file_pos = 0
             for line in in_file.readlines():
+                end_pos = file_pos + len(line)
                 if self.line_regex.search(line):
-                    count += 1
-        return count
-
-    def new(self, test_case, **kwargs):
-        return BinaryState.create(self.__count_instances(test_case))
-
-    def advance(self, test_case, state):
-        return state.advance()
-
-    def advance_on_success(self, test_case, state):
-        return state.advance_on_success(self.__count_instances(test_case))
-
-    def transform(self, test_case, state, process_event_notifier):
-        tmp = os.path.dirname(test_case)
-        with tempfile.NamedTemporaryFile(mode='w+', delete=False, dir=tmp) as tmp_file:
-            with open(test_case) as in_file:
-                i = 0
-                for line in in_file.readlines():
-                    if self.line_regex.search(line):
-                        if i < state.index or i >= state.end():
-                            tmp_file.write(line)
-                        i += 1
-                    else:
-                        tmp_file.write(line)
-
-        shutil.move(tmp_file.name, test_case)
-        return (PassResult.OK, state)
+                    hints.append({'p': [{'l': file_pos, 'r': end_pos}]})
+                file_pos = end_pos
+        return hints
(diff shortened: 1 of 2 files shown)


cvise/passes/line_markers.py
```diff
@@ -1,46 +1,21 @@
...
-import shutil
-import tempfile
 
-from cvise.passes.abstract import AbstractPass, BinaryState, PassResult
+from cvise.passes.hint_based import HintBasedPass
 
 
-class LineMarkersPass(AbstractPass):
+class LineMarkersPass(HintBasedPass):
     line_regex = re.compile('^\\s*#\\s*[0-9]+')
```
VitaNuo, line 7: Where does the logic that adds `#\s*[0-9]+` to the lines live and according to which principle?
I am missing the context a bit.

cvise/tests/test_line_markers.py
```diff
@@ -1,22 +1,23 @@
...
             tmp_file.write("# 1 'foo.h'\n# 2 'bar.h'\n#4   'x.h'")
 
-        state = self.pass_.new(tmp_file.name)
+        state = self.pass_.new(tmp_file.name, temp_dir=Path(self.temp_dir_))
         (_, state) = self.pass_.transform(tmp_file.name, state, None)
-        self.assertEqual(state.index, 0)
-        self.assertEqual(state.instances, 3)
 
         with open(tmp_file.name) as variant_file:
             variant = variant_file.read()
```
VitaNuo, line 23: Don't you need to assert that this variant is empty? AFAICT all lines match the regex.

cvise/tests/test_line_markers.py
```diff
@@ -28,11 +29,23 @@ def test_only_last(self):
         with tempfile.NamedTemporaryFile(mode='w', delete=False) as tmp_file:
             tmp_file.write("# 1 'foo.h'\n# 2 'bar.h'\n#4   'x.h\nint x = 2;")
 
-        state = self.pass_.new(tmp_file.name)
+        state = self.pass_.new(tmp_file.name, temp_dir=Path(self.temp_dir_))
```
VitaNuo, line 32: This is a little repetitive.. any chance it can be extracted into a helper function or turned into a parametrized test?

cvise/tests/test_line_markers.py
```diff
@@ -1,22 +1,23 @@
...
     def setUp(self):
+        self.temp_dir_ = self.enterContext(tempfile.TemporaryDirectory())
         self.pass_ = LineMarkersPass()
 
     def test_all(self):
         with tempfile.NamedTemporaryFile(mode='w', delete=False) as tmp_file:
             tmp_file.write("# 1 'foo.h'\n# 2 'bar.h'\n#4   'x.h'")
 
-        state = self.pass_.new(tmp_file.name)
+        state = self.pass_.new(tmp_file.name, temp_dir=Path(self.temp_dir_))
```
VitaNuo, line 19: Nit: `tmp` vs. `temp`.

cvise/passes/line_markers.py
emaxx-google: These are coming from a C preprocessor - hence typically show up in the minimization inputs.

GCC documents these line markers there: https://gcc.gnu.org/onlinedocs/gcc-3.0.2/cpp_9.html ; it also says `As an extension, the preprocessor accepts linemarkers in non-assembler input files`. For the purpose of minimization, they are mostly useless while still consuming a lot of space (typically 10-20% of the input file), so C-Reduce/C-Vise introduced this custom-tailred heuristic.

emaxx-google: I've added a documentation comment.

cvise/tests/test_line_markers.py
emaxx-google: done

cvise/tests/test_line_markers.py
emaxx-google: Yes that's what asserted below (line 26). Unless I misunderstood the question.

cvise/tests/test_line_markers.py
emaxx-google: Makes sense, but actually it's much simpler if I refactor tests - uploaded #188 (if accepted, I'll land that first and rebase the current PR on top).

cvise/passes/line_markers.py
VitaNuo: Cool, didn't know this.

Issue #166, opened 2025-01-23
topformflat confused by C++14/C23 digit separators


emaxx-google:
C++14 and C23 introduced single-quotation-mark as digit separators:

```c++
int x = 1'048'576;
```

The parser in `topformflat` doesn't understand this notation, misinterpreting this as a beginning of a string literal - which confuses all the parsing after that point. And this might lead to inserting invalid line breaks - for example, there:

```c++
int x = 1'000;
char c = '"';
int y = 1'000;
char* s = "some_##value";  // a line break gets added before `#` (it's misinterpreted as a preprocessor directive)
```

This impedes the performance of the `lines` passes on such programs, because the pass has to fall back to the non-reformatted file.

PR #191, opened 2025-06-25
hints: Migrate lines pass

emaxx-google:
This reimplements the Lines pass using the hint-based architecture.

Besides long-term improvements brought by the hint approach, this brings an actual functional improvement for this heuristic: no reformatting of input files is needed. Instead of overwriting the file in the "one block at a particular depth per line" fashion and deleting lines in this new text, we just emit hints for deletable code blocks.

This implies the following practical benefits:

1. The heuristic won't degrade dramatically when reformatting leads to a broken code (e.g., observed for C++14/C23 digit separators in #166);
2. Whitespaces aren't added/removed at various places in this pass;
3. Check_sanity is not needed anymore - we don't need to spend time checking whether the reformatted input still satisfies the interestingness test;
4. (minor) In C-Vise logs, the number of lines won't increase/decrease unpredictably whenever a Lines pass with some argument is started.

This also unblocks implementing the following improvements in the future:

5. Concurrently discovered reductions can be folded (merged) easily - there are no unrelated formatting changes that'd make merging very hard.
6. The Lines pass can run concurrently with other passes in an interleaving fashion, without having to maintain an extra input copy.

### DIFF
--- a/cvise/tests/test_lines.py
+++ b/cvise/tests/test_lines.py
@@ -61,7 +61,7 @@ def test_func_namespace_level1(tmp_path, input_path):
 
 def test_multiline_func_signature_level0(tmp_path, input_path):
     """Test that arg=0 deletes a top-level function despite line breaks in the signature."""
-    write_file(input_path, 'template <class T>\nint\nf()\n{\n}')
+    write_file(input_path, 'template <class T>\nSomeVeryLongType\nf()\n{\n}')
     p, state = init_pass('0', tmp_path, input_path)
     all_transforms = collect_all_transforms(p, state, input_path)
     assert '' in all_transforms
@@ -76,6 +76,8 @@ def test_multiline_func_signature_level1(tmp_path, input_path):
     all_transforms = collect_all_transforms(p, state, input_path)
     assert 'namespace {\n}\n' in all_transforms
     # the (multi-line) func must be deleted as a whole, not partially
+    # FIXME: Improve the heuristic to not try removing just the opening `namespace {` part,
+    # and replace the assertion here with "len(all_transforms) == 1".
     for s in all_transforms:
         assert ('template' in s) == ('f()' in s)
 
@@ -85,6 +87,7 @@ def test_c_comment(tmp_path, input_path):
     write_file(input_path, 'int x; /* \nsome\ncomment\n */\nint y;')
     p, state = init_pass('0', tmp_path, input_path)
     all_transforms = collect_all_transforms(p, state, input_path)
+    assert 'int x;' in all_transforms
     assert ' /* \nsome\ncomment\n */\nint y;' in all_transforms
     # no attempts to partially remove the comment
     for s in all_transforms:
@@ -96,10 +99,19 @@ def test_cpp_comment(tmp_path, input_path):
     write_file(input_path, 'int x; // some comment\nint y;')
     p, state = init_pass('0', tmp_path, input_path)
     all_transforms = collect_all_transforms(p, state, input_path)
+    assert 'int x;' in all_transforms
     assert ' // some comment\nint y;' in all_transforms
     # no attempts to partially remove the comment
     for s in all_transforms:
         assert ('//' in s) == ('some' in s) == ('comment' in s)
+
+
+def test_eof_with_non_recognized_chunk_end(tmp_path, input_path):
+    """Test the file terminating with a text that wouldn't be recognized as chunk end."""
+    write_file(input_path, '#define FOO }\nFOO')
+    p, state = init_pass('0', tmp_path, input_path)
+    all_transforms = collect_all_transforms(p, state, input_path)
+    assert '#define FOO }\n' in all_transforms
 
 
 def test_transform_deletes_lines_range(tmp_path, input_path):
@@ -123,10 +135,12 @@ def test_advance_on_success(tmp_path, input_path):
     """Test the scenario where successful advancements are interleaved with unsuccessful transforms."""
     write_file(input_path, 'foo;\nbar;\nbaz;\n')
     p, state = init_pass('0', tmp_path, input_path)
-    # cut 'foo' first
+    # Cut 'foo' first, pretending that all previous transforms (e.g., deletion of the whole text) didn't pass the
+    # interestingness test.
     state = advance_until(p, state, input_path, lambda s: 'bar' in s and 'baz' in s)
     p.advance_on_success(input_path, state)
-    # cut 'baz' now
+    # Cut 'baz' now, pretending that all transforms in between (e.g, deletion of "bar;") didn't pass the
+    # interestingness test.
     state = advance_until(p, state, input_path, lambda s: 'bar' in s)
     p.advance_on_success(input_path, state)
     assert read_file(input_path) == '\nbar;\n'
--- a/delta/topformflat.l
+++ b/delta/topformflat.l
@@ -27,7 +27,8 @@ int nesting = 0;
 // newlines are suppressed
 int threshold = 0;
 
-// update the position whenever a token is about to be processed.
+// update the position whenever a token is about to be processed
+// (this macro is triggered by Flex prior to executing the matched rule’s action).
 #define YY_USER_ACTION token_end_pos += yyleng;
 
 %}
@@ -56,7 +57,8 @@ int threshold = 0;
                 possibleChunkEnd(token_end_pos);
               }
 
-  /* a hash, then some non-newlines.  then, possibly, an escaped
+  /* parse a preprocessor directive:
+   * a hash, then some non-newlines.  then, possibly, an escaped
    * newline followed by more non-newlines (repeat as needed).
    * finally, a newline */
 "#".*("\\\n".*)*"\n" {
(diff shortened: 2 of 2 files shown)


cvise/tests/test_lines.py
```diff
@@ -38,108 +37,96 @@ def advance_until(pass_, state, input_path, predicate):
...
+    assert 'int f() {\n}\nnamespace foo\n{\nvoid g() {\n}\n}\n' in all_transforms
     # removal of code inside foo
-    assert 'int f() {\nchar x;\n}\nnamespace foo {\n}\n' in all_transforms
+    assert 'int f() {\nchar x;\n}\nnamespace foo\n{\n}\n' in all_transforms
 
 
-def test_multiline_func_signature_level0(input_path):
+def test_multiline_func_signature_level0(tmp_path, input_path):
     """Test that arg=0 deletes a top-level function despite line breaks in the signature."""
     write_file(input_path, 'template <class T>\nint\nf()\n{\n}')
```
VitaNuo, line 64: Nit: linebreak after the return type seems a little bizarre, but I guess that's intentional.

cvise/tests/test_lines.py
```diff
@@ -38,108 +37,96 @@ def advance_until(pass_, state, input_path, predicate):
...
     """Test that a C comment is deleted as a whole."""
     write_file(input_path, 'int x; /* \nsome\ncomment\n */\nint y;')
-    p, state = init_pass('0', input_path, check_sanity=lambda: True)
+    p, state = init_pass('0', tmp_path, input_path)
     all_transforms = collect_all_transforms(p, state, input_path)
-    assert 'int x;\nint y;\n' in all_transforms
+    assert ' /* \nsome\ncomment\n */\nint y;' in all_transforms
     # no attempts to partially remove the comment
     for s in all_transforms:
         assert ('/*' in s) == ('some' in s) == ('comment' in s) == ('*/' in s)
```
VitaNuo, line 91: Nit: you could also assert the other transforms for clarity.

```
assert 'int y;' in all_transforms
assert 'int x; /* \nsome\ncomment\n */' in all_transforms
```

Would this be correct?

cvise/tests/test_lines.py
```diff
@@ -38,108 +37,96 @@ def advance_until(pass_, state, input_path, predicate):
...
-    """Test the fallback scenario, when trimming spaces leads to an unsuccessful sanity check."""
-
-    def check_sanity():
-        if '  char' not in read_file(input_path):
-            raise InsaneTestCaseError([], '')
-
-    write_file(input_path, 'int f() {\n  char x;}\n')
-    p, _ = init_pass('1', input_path, check_sanity)
-    assert read_file(input_path) == 'int f() {\n   char x;\n}\n'
+    assert read_file(input_path) == '\nbar;\n'
```
VitaNuo, line 132: Same as above - this remaining newline at position 0 seems somewhat unusual.

delta/topformflat.l
```diff
@@ -124,9 +119,10 @@ int main(int argc, char *argv[])
 
   if (argc >= 2) {
     threshold = atoi(argv[1]);    // user-specified threshold
-  }  
+  }
 
   yyin = stdin;
   yylex();
+  possibleChunkEnd(token_end_pos);
```
VitaNuo, line 126: Is there some edge case where this is necessary? IMU every program will end with a `;` or a `}` anyways.

cvise/tests/test_lines.py
```diff
@@ -38,108 +37,96 @@ def advance_until(pass_, state, input_path, predicate):
...
     """Test that a C++ comment is deleted as a whole."""
     write_file(input_path, 'int x; // some comment\nint y;')
-    p, state = init_pass('0', input_path, check_sanity=lambda: True)
+    p, state = init_pass('0', tmp_path, input_path)
     all_transforms = collect_all_transforms(p, state, input_path)
-    assert 'int x;\nint y;\n' in all_transforms
+    assert ' // some comment\nint y;' in all_transforms
     # no attempts to partially remove the comment
     for s in all_transforms:
         assert ('//' in s) == ('some' in s) == ('comment' in s)
```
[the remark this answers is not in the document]
emaxx-google: Added, but it's the following assert that'd work here:

```python
assert 'int x;' in all_transforms
```

It's similar to the above - the new code doesn't care about making a comment a separate deletable entity; hence the comment happens to be attached to the subsequent deletable chunk - `int y` in this case. The old `topformflat` implementation would place the comment into a separate line.

cvise/tests/test_lines.py
```diff
@@ -38,108 +37,96 @@ def advance_until(pass_, state, input_path, predicate):
...
 
     This verifies the code performs the binary search or some similar strategy."""
     write_file(input_path, 'A;\nB;\nC;\nD;\nE;\nF;\nG;\nH;\n')
-    p, state = init_pass('0', input_path, check_sanity=lambda: True)
+    p, state = init_pass('0', tmp_path, input_path)
     all_transforms = collect_all_transforms(p, state, input_path)
     # deletion of a half:
     assert 'A;\nB;\nC;\nD;\n' in all_transforms
-    assert 'E;\nF;\nG;\nH;\n' in all_transforms
+    assert '\nE;\nF;\nG;\nH;\n' in all_transforms
```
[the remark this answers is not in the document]
emaxx-google: Before the PR, the implementation took an 8-line text and deleted the first 4 lines - which is why there was no leading line break.

After the PR, we stop caring closely about line breaks; `topformflat` sees `D;` as a deletable chunk and hence the hint ends up deleting `D;` while keeping the subsequent line break. It's not easy to avoid this - sometimes line breaks are crucial: like in this case where we're **not** allowed to delete `int y;<CRLF>`:

```c
int x; int y;
#define FOO
```

I'd also say keeping extra blank lines is harmless because deleting them is easy otherwise (there's also a separate pass `blank` for that).

cvise/tests/test_lines.py
```diff
@@ -38,108 +37,96 @@ def advance_until(pass_, state, input_path, predicate):
...
-    assert 'C;\nD;\nE;\nF;\nG;\nH;\n' in all_transforms
+    assert '\nC;\nD;\nE;\nF;\nG;\nH;\n' in all_transforms
     assert 'A;\nB;\nE;\nF;\nG;\nH;\n' in all_transforms
     assert 'A;\nB;\nC;\nD;\nG;\nH;\n' in all_transforms
     assert 'A;\nB;\nC;\nD;\nE;\nF;\n' in all_transforms
 
 
-def test_advance_on_success(input_path):
+def test_advance_on_success(tmp_path, input_path):
     """Test the scenario where successful advancements are interleaved with unsuccessful transforms."""
```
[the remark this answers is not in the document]
emaxx-google: I've added comments to explain what's happening here - but to paraphrase it:

1. Deleting all `foo;\nbar;\nbaz;\n` - pretend that it wasn't accepted by the interestingness test.
2. Deleting `foo;` - pretend that it was accepted.
3. Deleting `bar;` - pretend that it wasn't accepted.
4. Deleting `baz;` - pretend that it was accepted.

cvise/tests/test_lines.py
emaxx-google: It's due to the same underlying root cause, so let's continue the discussion above.

delta/topformflat.l
```diff
@@ -24,6 +27,9 @@ int nesting = 0;
 // newlines are suppressed
 int threshold = 0;
 
+// update the position whenever a token is about to be processed.
+#define YY_USER_ACTION token_end_pos += yyleng;
```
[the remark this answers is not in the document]
emaxx-google: Added a comment - there's magic in Flex that triggers this macro: https://westes.github.io/flex/manual/Misc-Macros.html.

delta/topformflat.l
```diff
@@ -38,81 +44,70 @@ int threshold = 0;
...
-"\'"          { diag("<CHAR>"); emit(); BEGIN(CHARLIT); }   /* start tick */
+"\'"          { diag("<CHAR>"); BEGIN(CHARLIT); }           /* start tick */
 
 <CHARLIT>{
-  "\\"(.|\n)  { emit(); }                                   /* escaped character */
-  "\'"        { emit(); diag("</CHAR>"); BEGIN(INITIAL); }  /* close tick */
-  (.|\n)      { emit(); }                                   /* ordinary char */
+  "\\"(.|\n)  { }                                           /* escaped character */
+  "\'"        { diag("</CHAR>"); BEGIN(INITIAL); }          /* close tick */
+  (.|\n)      {  }                                          /* ordinary char */
```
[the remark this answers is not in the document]
emaxx-google: Just remove all spaces except the two ones, like how `clang-format` typically does it? The original code aligned all comments vertically, which is what I tried to preserve with the changes.

emaxx-google (commented):
Thanks!

cvise/tests/test_lines.py
VitaNuo: Thanks.

delta/topformflat.l
VitaNuo: Thanks!

cvise/tests/test_lines.py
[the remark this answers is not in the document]
VitaNuo: Makes sense, thanks for the explanation.

PR #192, opened 2025-06-25
Improve Lines: only remove exactly at given depth

emaxx-google:
Stop producing hints for deleting code at levels smaller than the specified
depth (threshold). This should significantly improve the success rate for
all depths greater than zero, because we'll stop trying partially removing
outer constructs.

For example, for a top-level C function, running Lines with arg=1 should
only attempt removing statements inside the function body, but don't
try deleting the function signature alone or deleting the closing curly
brace alone.

Note: the downside of this change is that when running Lines for arg>1,
the heuristic will stop trying to delete higher-level chunks of code. We
consider this not to be a problem because one of planned follow-up
changes will run all Lines passes concurrently, in an interleaving fashion.

### DIFF
--- a/cvise/tests/test_lines.py
+++ b/cvise/tests/test_lines.py
@@ -37,15 +37,31 @@ def advance_until(pass_, state, input_path, predicate):
         assert state is not None
 
 
+def is_valid_brace_sequence(s):
+    balance = 0
+    for c in s:
+        if c == '{':
+            balance += 1
+        elif c == '}':
+            balance -= 1
+        if balance < 0:
+            return False
+    return balance == 0
+
+
 def test_func_namespace_level0(tmp_path, input_path):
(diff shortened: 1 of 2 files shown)


cvise/tests/test_lines.py
```diff
@@ -37,49 +37,238 @@ def advance_until(pass_, state, input_path, predicate):
...
+    # check no transform violates curly brace sequences
+    for s in all_transforms:
+        assert is_valid_brace_sequence(s)
+
+
+def test_class_with_methods_level0(tmp_path, input_path):
+    """Test that arg=0 deletes the whole class definition."""
+    write_file(
+        input_path,
+        """
```
emaxx-google, line 119: How do you find this format of tests? Now that the new hint-based implementation stops juggling whitespaces, we can rewrite all tests in this style.

PR #193, opened 2025-06-26
hints: Switch to the zstandard library for compression

emaxx-google:
The Zstandard Python package seems to have a better compatibility than Pyzstd at the moment.

Also add a test that verifies a nontrivial compression actually takes place.


VitaNuo (approved):
Thanks!

PR #195, opened 2025-06-27
Refactor job management in TestManager

emaxx-google:
Store all per-job information in a separate Job object instead of using several independent dicts. Also explicitly store the Pass in the Job, instead of assuming all active jobs have the same pass.

This is preparation for the introduction of interleaved (concurrent) execution of different passes. The current PR is intended to have no functional changes.


VitaNuo (approved):
Thanks!

emaxx-google:
thank you for the review.

PR #202, opened 2025-07-03
hints: Migrate clang_delta logic for RemoveUnusedFunction

emaxx-google:
This adds a new parameter to clang_delta: --generate-hints, which results in emitting JSONL hints instead of the modified text. With this commit, this mode is only supported for the
RemoveUnusedFunction transformation.

Follow-up changes will wire up the Python "pass" logic with this new functionality.

### DIFF
--- a/clang_delta/HintsBuilder.cpp
+++ b/clang_delta/HintsBuilder.cpp
@@ -1,0 +1,65 @@
+#include "HintsBuilder.h"
+
+#include <stdint.h>
+
+#include <algorithm>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "clang/Basic/LangOptions.h"
+#include "clang/Basic/SourceLocation.h"
+#include "clang/Basic/SourceManager.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/FormatVariadic.h"
+
+HintsBuilder::HintsBuilder(clang::SourceManager &SM, const clang::LangOptions &LO)
+: SourceMgr(SM), NoOpRewriter(SourceMgr, LO) {}
+
+HintsBuilder::~HintsBuilder() = default;
+
+void HintsBuilder::AddPatch(clang::SourceRange R) {
+  AddPatch(R.getBegin(), NoOpRewriter.getRangeSize(R));
+}
+
+void HintsBuilder::AddPatch(clang::CharSourceRange R) {
+  AddPatch(R.getBegin(), NoOpRewriter.getRangeSize(R));
+}
+
+void HintsBuilder::AddPatch(clang::SourceLocation L, int64_t Len) {
+  if (Len <= 0) {
+    // This would be an invalid hint patch.
+    return;
+  }
+  Patch P;
+  P.L = SourceMgr.getFileOffset(L);
+  P.R = P.L + Len;
+  CurrentHint.Patches.push_back(P);
+}
+
+void HintsBuilder::FinishCurrentHint() {
+  if (CurrentHint.Patches.empty()) {
+    // Don't add empty hints.
+    return;
+  }
+  Hints.push_back(std::move(CurrentHint));
+  CurrentHint = {};
+}
+
+void HintsBuilder::ReverseOrder() {
+  std::reverse(Hints.begin(), Hints.end());
+}
+
+std::vector<std::string> HintsBuilder::GetHintJsons() const {
+  std::vector<std::string> Jsons;
+  for (const auto& H : Hints) {
+    std::string Array;
+    for (const auto& P : H.Patches) {
+      if (!Array.empty())
+        Array += ",";
+      Array += llvm::formatv(R"txt({{"l":{0},"r":{1}})txt", P.L, P.R);
+    }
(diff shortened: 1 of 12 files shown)


clang_delta/RemoveUnusedFunction.cpp
```diff
@@ -508,8 +517,11 @@ void RemoveUnusedFunction::removeOneFunctionDecl(const FunctionDecl *FD)
 
     LocEnd = getFunctionLocEnd(FuncLocStart, LocEnd, FD);
     if (SrcManager->isWrittenInMainFile(FuncLocStart) &&
-        SrcManager->isWrittenInMainFile(LocEnd))
-      TheRewriter.RemoveText(SourceRange(FuncLocStart, LocEnd));
+        SrcManager->isWrittenInMainFile(LocEnd)) {
+      SourceRange R(FuncLocStart, LocEnd);
+      Hints->AddPatch(R);
+      TheRewriter.RemoveText(R);
```
emaxx-google, line 523: I'm planning to eventually delete these old invocations of the `TheRewriter` , but only after the Python counterpart is migrated to use hints. Or maybe we'll keep them longer in case someone uses clang_delta for other outside C-Vise (but I'm not aware of any, and it's not a public API).

clang_delta/tests/test_clang_delta.py
```diff
@@ -588,105 +630,216 @@ def test_remove_unused_function_class(self):
             'remove-unused-function/class.cc',
             '--transformation=remove-unused-function --counter=1',
         )
```
emaxx-google, line 632: Ditto with old test assertions for `remove-unused-function` - we can delete them once we switch everything to hints, which is why I didn't invest into any clever way of parameterizing the tests for "with hints and w/o hints".

clang_delta/RemoveUnusedFunction.cpp
```diff
@@ -328,18 +328,27 @@ void RemoveUnusedFunction::doRewriting()
...
+  if (ToCounter == std::numeric_limits<int>::max()) {
+    // This special value denotes performing all possible transforms.
+    ToCounter = static_cast<int>(AllValidFunctionDecls.size());
+  } else {
+    TransAssert((TransformationCounter <=
+                  static_cast<int>(AllValidFunctionDecls.size())) &&
+                "TransformationCounter is larger than the number of defs!");
+    TransAssert((ToCounter <= static_cast<int>(AllValidFunctionDecls.size())) &&
+                "ToCounter is larger than the number of defs!");
+  }
```
VitaNuo, line 340: Nit: at least for the second assert, it seems like it could be refactored out of the `if..else`, since `ToCounter <= static_cast<int>(AllValidFunctionDecls.size())` is true after `ToCounter = static_cast<int>(AllValidFunctionDecls.size());`. 

Maybe the first one as well?

clang_delta/HintsBuilder.cpp
```diff
@@ -0,0 +1,65 @@
...
+    return;
+  }
+  Patch P;
+  P.L = SourceMgr.getFileOffset(L);
+  P.R = P.L + Len;
+  CurrentHint.Patches.push_back(P);
+}
+
+void HintsBuilder::FinishCurrentHint() {
+  if (CurrentHint.Patches.empty()) {
```
VitaNuo, line 41: Nit: can this actually happen?

clang_delta/RemoveUnusedFunction.cpp
VitaNuo: SG!

clang_delta/TransformationManager.cpp
```diff
@@ -320,7 +320,17 @@ bool TransformationManager::doTransformation(std::string &ErrorMsg, int &ErrorCo
 
   llvm::raw_ostream *OutStream = getOutStream();
   bool RV;
+  if (GenerateHints) {
   if (CurrentTransformationImpl->transSuccess()) {
```
VitaNuo, line 324: Nit: indentation.

clang_delta/TransformationManager.cpp
```diff
@@ -320,7 +320,17 @@ bool TransformationManager::doTransformation(std::string &ErrorMsg, int &ErrorCo
 
   llvm::raw_ostream *OutStream = getOutStream();
   bool RV;
```
VitaNuo, line 322: Nit: you can set this to default to `true` and only change to `false` wherever necessary.

clang_delta/TransformationManager.cpp
```diff
@@ -320,7 +320,17 @@ bool TransformationManager::doTransformation(std::string &ErrorMsg, int &ErrorCo
...
+  if (GenerateHints) {
   if (CurrentTransformationImpl->transSuccess()) {
+      CurrentTransformationImpl->outputHints(*OutStream);
+      RV = true;
+    }
+    else {
+      CurrentTransformationImpl->getTransErrorMsg(ErrorMsg);
```
VitaNuo, line 329: Don't you need to call this in non-generate-hints case as well?

VitaNuo (approved):
Thanks! Left  a few nits.

PR #203, opened 2025-07-03
hints: Add ClangHintsPass; start from RemoveUnusedFunction

emaxx-google:
Implement the Python counterpart of the hint-based clang_delta logic added by previous patches.

The direct improvement is that we don't need to build Clang AST for every function deletion attempt anymore - instead, we essentially precalculate all transforms (that's what "hints" are) and can quickly apply them via pure text editing.

Additionally, the migration will allow running this pass in an interleaving fashion with other passes, and folding concurrently discovered reductions.


cvise/passes/clanghints.py
```diff
@@ -0,0 +1,92 @@
...
+    we want to brute-force Clang's `--std=` parameter that maximizes the generated set of hints. This requires having
+    special logic in new() and carrying over some extra information from new() to advance_on_success() throughout all
+    advance() calls.
+    """
+
+    def check_prerequisites(self):
+        return self.check_external_program('clang_delta')
+
+    def new(self, test_case, tmp_dir, **kwargs):
+        # Choose the best standard unless the user provided one.
```
emaxx-google, line 45: A note for reviewing: conceptually this follows the code from `clangbinarysearch.py`.

cvise/passes/clanghints.py
```diff
@@ -0,0 +1,92 @@
...
+            f'--transformation={self.arg}',
+            f'--std={std}',
+            '--generate-hints',
+            test_case,
+        ]
+
+        logging.debug(shlex.join(str(s) for s in cmd))
+
+        hints = []
+        with subprocess.Popen(cmd, stdout=subprocess.PIPE, text=True) as proc:
```
emaxx-google, line 88: One missing thing, compared to `clangbinarysearch.py`, is that we don't set a timeout here. It's crucial for the cases when building Clang AST hangs.

I'm planning to address this in a follow-up - I want to implement it in a more robust way than the original code's hardcoded 10 seconds (https://github.com/marxin/cvise/blob/ed1cba72a7c3227841527e73cdf4f15487c09241/cvise/passes/clangbinarysearch.py#L13).

VitaNuo (approved):
Thanks!

emaxx-google:
Added tests too.

emaxx-google:
Thanks for reviewing.

PR #204, opened 2025-07-04
Fix --stopping-threshold checked too rarely

emaxx-google:
There was a bug that the threshold check kept repeatedly using the same file size - the one before the loop over passes started.

This is probably a rarely used feature, but uploading this as a separate fix to make follow-up refactoring of the same code easier to review.


VitaNuo (approved):
Thanks!

PR #205, opened 2025-07-04
Refactor pass category logic of pass group config

emaxx-google:
This makes the difference between pass categories ("first"/"main"/"last") hardcoded in a more declarative way in the code, and unifies the logic of how their passes are executed.

It's intended to be a pure refactoring change. It's a preparation for the introduction of "interleaving" pass categories, for which we want to reuse most of the existing logic.


VitaNuo (approved):
Thanks!

emaxx-google (commented):
thanks!

PR #207, opened 2025-07-04
interleaving: Add support into pass group JSON config

emaxx-google:
Add "interleaving" and "interleaving_first" pass categories into the pass_group config parser. The passes in these categories will run in an interleaving fashion (concurrently).

Another change made here is making all pass categories optional - this is to allow the user play easier with various combinations of these strategies.

### DIFF
--- a/tests/test_cvise.py
+++ b/tests/test_cvise.py
@@ -96,14 +96,15 @@ def test_interleaving_passes(tmp_path: Path):
     shutil.copy(testcase_path, '.')
     copy_path = Path(testcase_path.name)
 
-    proc = start_cvise(
-        ['-c', 'gcc -c test.c && grep foo test.c', '--pass-group-file', config_path, testcase_path.name]
-    )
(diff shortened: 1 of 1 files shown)


cvise/cvise.py
```diff
@@ -44,15 +44,26 @@ class PassCategory:
...
+            log_title='INITIAL PASSES (interleaving)',
+            initial=True,
+            interleaving=True,
+            once=True,
+        ),
+        PassCategory(name='first', log_title='INITIAL PASSES', initial=True, interleaving=False, once=True),
         # "main" categories (looped)
-        PassCategory(name='main', log_title='MAIN PASSES', initial=False, once=False),
+        PassCategory(
+            name='interleaving', log_title='MAIN PASSES (interleaving)', initial=False, interleaving=True, once=False
```
VitaNuo, line 62: Nit: why not `INTERLEAVING PASSES`?

VitaNuo (approved):

PR #208, opened 2025-07-07
hints: Add bundle class

emaxx-google:
Refactor the code to pass hints as a bundle object instead of the naked list.

This is preparation for introducing additional items of data stored next to hints - e.g., "vocabulary" of strings shared between hints.

### DIFF
--- a/cvise/utils/hint.py
+++ b/cvise/utils/hint.py
@@ -12,7 +12,7 @@ applied to all heuristics in a uniform way).
 from dataclasses import dataclass
 import json
 from pathlib import Path
-from typing import Any, List, Sequence, TextIO
+from typing import List, Sequence
 import zstandard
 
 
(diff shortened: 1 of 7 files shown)


cvise/utils/hint.py
```diff
@@ -48,9 +65,9 @@
 }
 
 
-def apply_hints(hints: Sequence[object], file: Path) -> str:
+def apply_hints(hints: HintBundle, file: Path) -> str:
```
VitaNuo, line 68: Nit: consider calling the variable `bundle` instead of `hints` to avoid looping over `hints.hints`, etc.

VitaNuo (approved):
Thanks!

cvise/utils/hint.py
emaxx-google: Done, makes sense.

PR #210, opened 2025-07-07
hints: Migrate clang_delta ReplaceFunctionDefWithDecl

emaxx-google:
Migrate the ReplaceFunctionDefWithDecl heuristic to the hint-based architecture, which implies switching it to the "clanghints" pass.

As with other changes in this effort, this removes the need in building the Clang AST for every transformation attempt, and paves the way for better parallelization during reduction.


emaxx-google (commented):
thanks!

emaxx-google:
@VitaNuo Github somehow lost your additional comments sent after the review, so replying here:

> Nit: if you're following LLVM style, then these curly braces are excessive.

Done. (I'm adding them subconsciously - too used to the Chromium style guide :) )

> Why not *It?

Not sure I understand the proposal. `*It` would return the string in the vocabulary - what we need is an index.

VitaNuo:
> Not sure I understand the proposal. *It would return the string in the vocabulary - what we need is an index.
Yes, right. I got confused, sorry. The patch LGTM.

PR #212, opened 2025-07-08
Stop using check_sanity for passes

emaxx-google:
Since the migration of the Lines pass to hints, none of heuristics modifies the input file during initialization. This means the "check_sanity" logic (double-checking whether the modified input still satisfies the interestingness script) isn't needed anymore as a pass parameter.


emaxx-google:
Thanks.

PR #216, opened 2025-07-08
CI: Don't cancel jobs for the master branch

emaxx-google:
Even when multiple commits land in a quick succession to the master branch, we still want to test them all; cancelling concurrent jobs was resulting in spurious "failures" displayed on Github.

### DIFF
--- a/.github/workflows/build-llvm-nightly.yml
+++ b/.github/workflows/build-llvm-nightly.yml
@@ -9,9 +9,10 @@ on:
     - cron: '0 3 * * *'
   workflow_dispatch: {}
 
+# Cancel the workflow if a new one is requested for the same PR/branch, but still test all commits on the master branch.
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master'}}
 
 jobs:
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,9 +7,10 @@ on:
     branches: [ master ]
   workflow_dispatch: {}
 
+# Cancel the workflow if a new one is requested for the same PR/branch, but still test all commits on the master branch.
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -7,9 +7,10 @@ on:
     branches: [ master ]
   workflow_dispatch: {}
 
+# Cancel the workflow if a new one is requested for the same PR/branch, but still test all commits on the master branch.
```
VitaNuo, line 10: Is this statement correct though? AFAIU all the jobs that are not testing a commit on master will be canceled, not only the job for the same PR.

VitaNuo (commented):
Thanks!

.github/workflows/build.yml
VitaNuo: Ah ok, so you would like to allow canceling in-progress jobs for PRs, if multiple commits land in quick succession, but never cancel on master (rather queue up). Ok thanks, got it now.

VitaNuo (approved):
Thanks!

.github/workflows/build-llvm-nightly.yml
```diff
@@ -9,9 +9,10 @@ on:
     - cron: '0 3 * * *'
   workflow_dispatch: {}
 
+# Cancel the workflow if a new one is requested for the same PR/branch, but still test all commits on the master branch.
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master'}}
```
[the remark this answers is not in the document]
emaxx-google: There also are `push:` and `pull_request:` properties there - and as the docs say it: "If you specify multiple events, only one of those events needs to occur to trigger your workflow" https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-multiple-events . So when one lands multiple commits quickly, these `push` events triggered multiple builds, all of which except one would be canceled. The latter part is what the PR aims to change.

.github/workflows/build.yml
emaxx-google: Yup, exactly. Thanks for the review!

PR #219, opened 2025-07-09
hints: Introduce hint type; run independent binsrch per type

emaxx-google:
This will be used for passes that return several groups of reduction opportunities, with different success rate and whose contiguous ranges are better to attempt separately (e.g., the "balanced" pass replacing round bracket pairs vs. square bracket pairs).


VitaNuo (approved):
Thanks!

PR #220, opened 2025-07-09
Add a variable for local tools directory.

VitaNuo:
This allows to set a flexible path instead of relying to on the fragile directory layout, referring to the parent directory with `..`, etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,6 +128,7 @@ set(cvise_PACKAGE_URL         "https://github.com/marxin/cvise/")
 set(cvise_PACKAGE_VERSION     "2.11.0")
 set(cvise_VERSION             "2.11.0")
 set(cvise_LLVM_VERSION        "${LLVM_PACKAGE_VERSION}")
+set(LOCAL_TOOLS_DIR,          "${CMAKE_SOURCE_DIR}")
 
 configure_file("cmake_config.h.in" "${PROJECT_BINARY_DIR}/config.h")
 add_definitions("-DHAVE_CONFIG_H")
--- a/cvise/utils/externalprograms.py
+++ b/cvise/utils/externalprograms.py
@@ -18,7 +18,7 @@ def find_external_programs():
     for prog, local_folder in programs.items():
         path = None
         if local_folder:
-            local_folder = os.path.join(SCRIPT_PATH, '..', '..', local_folder)
+            local_folder = os.path.join('@LOCAL_TOOLS_DIR@', local_folder)
             if platform.system() == 'Windows':
                 for configuration in ['Debug', 'Release']:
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -128,6 +128,7 @@ set(cvise_PACKAGE_URL         "https://github.com/marxin/cvise/")
 set(cvise_PACKAGE_VERSION     "2.11.0")
 set(cvise_VERSION             "2.11.0")
 set(cvise_LLVM_VERSION        "${LLVM_PACKAGE_VERSION}")
+set(LOCAL_TOOLS_DIR,          "${CMAKE_SOURCE_DIR}")
```
emaxx-google, line 131: I would've expected this to be `PROJECT_BINARY_DIR`, but ultimately unit tests should show.

VitaNuo: Tests pass. Would you like me to try with `PROJECT_BINARY_DIR` as well?

emaxx-google: Hmm, in my local checkout of this branch I get the following error in tests: `FileNotFoundError: [Errno 2] No such file or directory: 'delta'`. I'm not sure how it works on bots.

emaxx-google: Bots seem to work regardless of this code, because the CI action `cd`s into the build directory: https://github.com/marxin/cvise/blob/f3e563c470bbba7d83262bd647c27ef788c36726/.github/workflows/build.yml#L57 . So the fallback `shutil.which(prog)` always finds something in the bot environment, meanwhile running the tests as I do it locally - `pytest build` - doesn't.

CMakeLists.txt
```diff
@@ -128,6 +128,7 @@ set(cvise_PACKAGE_URL         "https://github.com/marxin/cvise/")
 set(cvise_PACKAGE_VERSION     "2.11.0")
 set(cvise_VERSION             "2.11.0")
 set(cvise_LLVM_VERSION        "${LLVM_PACKAGE_VERSION}")
+set(LOCAL_TOOLS_DIR          "${PROJECT_BINARY_DIR}")
```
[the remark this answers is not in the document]
VitaNuo: Adjusted the alignment.

cvise/utils/externalprograms.py
```diff
@@ -18,7 +18,7 @@ def find_external_programs():
     for prog, local_folder in programs.items():
         path = None
         if local_folder:
-            local_folder = os.path.join(SCRIPT_PATH, '..', '..', local_folder)
```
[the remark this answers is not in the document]
VitaNuo: Changed approach as discussed offline, so this variable is still needed.

VitaNuo:
Thanks, updated as discussed.

PR #222, opened 2025-07-09
lines: Configure Flex to use fast scanning mode

emaxx-google:
"%option fast" makes the topformflat scanner faster at the cost of binary size. Empirical improvement is ~50%: improving the throughput from 200 MiB/sec to 300 MiB/sec (numbers taken on my machine). The topformflat binary size increase is O(KBs), which is negligible.

### DIFF
--- a/delta/topformflat.l
+++ b/delta/topformflat.l
@@ -34,6 +34,8 @@ int threshold = 0;
 
 %}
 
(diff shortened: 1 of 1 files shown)


VitaNuo (approved):
Thanks!

PR #226, opened 2025-07-10
hints: Migrate blank pass

emaxx-google:
Similar to other commits in the series, porting a pass to the hint-based architecture allows for better parallelization in the future.

I've ported the heuristic as-is - with both '^\s*$' and '^#' regexps, even though I'm not sure about the benefit of the second pattern. The new implementation gives them different "t", so that they're attempted independently, like the old implementation was doing.


VitaNuo (approved):
Thanks!

PR #227, opened 2025-07-10
hints: Migrate comments pass

emaxx-google:
Similar to other commits in the series, porting a pass to the hint-based architecture allows for better parallelization in the future.

I've ported the heuristic as-is - even though it being a simple regexp doesn't handle many cases correctly, like `foo("http://example.com")` being mistreated as a comment because of the double-slash.

### DIFF
--- a/.github/workflows/build-llvm-nightly.yml
+++ b/.github/workflows/build-llvm-nightly.yml
@@ -9,9 +9,10 @@ on:
     - cron: '0 3 * * *'
   workflow_dispatch: {}
 
+# Cancel the workflow if a new one is requested for the same PR/branch, but still test all commits on the master branch.
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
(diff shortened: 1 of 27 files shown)


cvise/passes/comments.py
```diff
@@ -1,42 +1,26 @@
...
-            if prog != prog2:
-                with open(test_case, 'w') as out_file:
-                    out_file.write(prog2)
 
-                return (PassResult.OK, state)
-            else:
-                state = self.advance(test_case, state)
+        hints = []
+        # Remove all multiline comments
+        # Replace /* any number of * if not followed by / or anything but * */
```
VitaNuo, line 17: Could you rephrase this comment? It's hard to understand what this is saying.
The regex below is also fairly incomprehensible (I see it's the same one as before), so it'd be good to have a decent comment at least.

VitaNuo (approved):
Thanks!

cvise/passes/comments.py
emaxx-google: Sure, done.

PR #229, opened 2025-07-10
clang_delta: Use real timeout for old&new passes

emaxx-google:
The old ClangBinarySearchPass class used to hardcode the 10-second timeout, which was insufficient for big inputs and led to a significant slowdown of the reduction as this heuristic didn't work at all. The new ClangHintsPass didn't have the timeout logic at all, which would result in hanging C-Vise if building Clang AST hangs.

Fix both places to use the --timeout command-line parameter value.


VitaNuo (approved):
Thanks!

emaxx-google:
Thank you for the reviews!

PR #231, opened 2025-07-10
hints: Fix type annotations

emaxx-google:
Fix some trivial violations. Also fix HintState to not store empty per-type substates - to make the type linters happy but also to have a better preconditions in the code.

Found using a manual execution of Mypy (a TODO would be to properly integrate it into CI).


emaxx-google:
There was a bug in the previous patch - fixed it now and covered with tests.

emaxx-google:
Thanks for the review.

PR #232, opened 2025-07-10
hints: Fix 't' incorrect place for Blank pass and in Schema

emaxx-google:
The hint type property should be specified on the hint object, not on the patch. This was incorrectly documented in the Schema and incorrectly used in the Blank pass.

Fix both places and add tests with strict sanity checks.


emaxx-google (commented):
Thanks for the review!

emaxx-google:
Also came up with another regression test.

PR #238, opened 2025-07-19
interleaving: Run all Pass.new() concurrently

emaxx-google:
Initialize all passes simultaneously, on the process pool, when the interleaving mode is used.

This makes the interleaving mode closer to be practically useful, since the majority of time was spent on running all new() methods sequentially - which was worse than the old single-pass mode.


VitaNuo (approved):
Thanks!

PR #239, opened 2025-07-22
interleaving: Support applying multiple hint bundles

emaxx-google:
With this commit, the code that applies given hints to a file can accept multiple hint bundles.

This is preparation for implementing folding: attempts to merge successful discoveries made by different passes concurrently.


VitaNuo (approved):
Could you please see what's up with the failing check? Thanks!

PR #240, opened 2025-07-22
test_manager: Further decouple job scheduling

emaxx-google:
Refactor the task loop to move all job-specific checks (which passes are initialized, etc.) into corresponding helpers.

This is preparation for introducing a new type of jobs: folding (i.e., merging the reductions concurrently discovered by different passes).


VitaNuo (approved):

emaxx-google:
thanks for the reviews

PR #243, opened 2025-07-22
test_manager: Proceed on any success, not necessarily the first

emaxx-google:
Pick an arbitrary transformation among multiple ones that succeeded, instead of preferring the one that's scheduled earlier.

The main goal is the upcoming support of folding (i.e., merging multiple concurrently discovered successful transformations), where preferring a bigger, rather than an earlier, reduction will be crucial. Other benefits: slow/timing-out jobs won't "stand in the way" of proceeding with a successful transformation; implementation of TestManager doesn't need to repeat the job status checking logic in two different places.

The downside of this change is that it makes reductions non-deterministic. Additionally, there's a temporary problem that C-Vise might skip some opportunities in the "cleanup" passes - for example, to avoid the flakiness in test_cvise.py we had to mention the "blank" pass twice; we plan to fix it in follow-ups.

### DIFF
--- a/cvise/pass_groups/all.json
+++ b/cvise/pass_groups/all.json
@@ -189,6 +189,7 @@
     {"pass": "clex", "arg": "rename-toks", "renaming": true},
     {"pass": "clex", "arg": "delete-string"},
     {"pass": "blank"},
+    {"pass": "blank"},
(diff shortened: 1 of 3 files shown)


emaxx-google:
thank you!

Issue #59, opened 2021-07-23
`IncludesPass` and `CommentsPass` have issues with files using "non-standard" encodings


andrewvaughanj:
For:

```
cat unit.c
/*
* für
*/ 
```

```
file unit.c
unit.c: ISO-8859 text
```

running `cvise --start-with-pass IncludesPass` gives:

```
00:00:00 INFO ===< 118085 >===
00:00:00 INFO running 24 interestingness tests in parallel
00:00:00 INFO INITIAL PASSES
00:00:00 INFO ===< IncludesPass >===
00:00:00 WARNING IncludesPass has encountered a non fatal bug: pass got stuck
00:00:00 INFO ===< UnIfDefPass >===
00:00:00 INFO ===< CommentsPass >===
00:00:00 WARNING CommentsPass has encountered a non fatal bug: pass got stuck
00:00:00 INFO ===< IfPass >===
Traceback (most recent call last):
  File "/usr/bin/cvise", line 286, in <module>
    reducer.reduce(pass_group, skip_initial=args.skip_initial_passes)
  File "/usr/share/cvise/cvise.py", line 143, in reduce
    self._run_additional_passes(pass_group['first'])
  File "/usr/share/cvise/cvise.py", line 166, in _run_additional_passes
    self.test_manager.run_pass(p)
  File "/usr/share/cvise/utils/testing.py", line 503, in run_pass
    self.state = self.current_pass.new(self.current_test_case, self.check_sanity)
  File "/usr/share/cvise/passes/ifs.py", line 36, in new
    bs = BinaryState.create(self.__count_instances(test_case))
  File "/usr/share/cvise/passes/ifs.py", line 22, in __count_instances
    for line in in_file.readlines():
  File "/usr/lib64/python3.8/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfc in position 6: invalid start byte
```

(this was after editing `cvise/utils/testing.py`  to set `GIVEUP_CONSTANT` to 1)

and creates:

```
Package: cvise 2.3.0
Git version: unknown
LLVM version: 12.0.0
System: uname_result(system='Linux', node='vistrrdslin0001', release='5.12.13-1-default', version='#1 SMP Mon Jun 28 06:37:23 UTC 2021 (74bd8c0)', machine='x86_64', processor='x86_64')
***************************************************

IncludesPass has encountered a bug:
[...]

PR #244, opened 2025-07-23
Support non UTF-8 inputs

emaxx-google:
Accept test cases that aren't valid UTF-8. While unusual in modern programming, non-UTF-8 files do still occur (e.g., in tests of text parsing libraries).

This commit adds support of non-UTF-8 reductions into the following passes: blank, comments, clanghints, line_markers, lines. Other passes will be silently skipped on Unicode errors. Also make the "--print-diff" logic support non-UTF-8 files too.

This should address most of #59.

### DIFF
--- a/cvise-cli.py
+++ b/cvise-cli.py
@@ -476,43 +476,47 @@ def do_reduce(args):
         print(err)
     else:
         time_stop = time.monotonic()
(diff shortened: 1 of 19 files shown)


VitaNuo (approved):
Thank you!

PR #246, opened 2025-07-25
hints: Add preamble to the bundle file

emaxx-google:
It's intended to store properties common to all hints. (The specific example will be added in a follow-up change.)

Additionally, the preamble can act as a self-documentation and can allow us make backwards-incompatible format changes without having to update all heuristics.


VitaNuo (approved):
Thank you!

PR #247, opened 2025-07-25
interleaving: Report succeeding pass(es)

emaxx-google:
Log name(s) of pass(es) that gave the successful transformation.

Implementation-wise, this is less trivial for folds, since we need to remember where each hint originates from, and carry this information through when merging overlapping hints. The result is a bit imprecise when overlapping-but-not-contained hints are involved, but since this is information-only we consider this OK.


VitaNuo (approved):
Thank you!

PR #248, opened 2025-07-25
Bump nightly CI to LLVM 22

marxin:


emaxx-google:
> I couldn't come up with a simple workaround on the Github Action side for this (and something that wouldn't break again once the script gets updated). Probably the ideal solution would be if the upstream `llvm.sh` would provide an option "register the latest Debian repository and install the latest version available in it" instead of having to hardcode the version in 2 different places.

I uploaded https://github.com/opencollab/llvm-jenkins.debian.net/pull/34, attempting to propose the latter idea. Waiting for reviews.

marxin:
Apparently, there will be multiple changes related to LLVM 22. I am going to work on that later this week.

emaxx-google:
There are large-scale refactorings and AST changes happening in Clang tip-of-the-tree right now, and as far as I can see they haven't finished yet - probably it'll make sense to hold off adopting them in C-Vise until the dust settles. I suspect we'll need more tests for clang_delta to go through this, because of the scale of changes I've seen needed in other Clang-based tools right now...

PR #249, opened 2025-07-25
interleaving: Enable by default

emaxx-google:
Add an interleaving category into the all.json config - this means C-Vise will use the new logic with running different passes concurrently by default. In case of problems or performance issues with it, the user can fall back to "--pass-group no-interleaving" (this config is the old all.json before this commit).

For now, we only use the "interleaving_first" category, which means going through all reduction candidates once - without looping until all passes are exhausted. We'll reevaluate this later once more performance improvements get into the "interleaving" code path and more passes are migrated to be compatible with it.

### DIFF
--- a/cvise/CMakeLists.txt
+++ b/cvise/CMakeLists.txt
@@ -65,6 +65,7 @@ endfunction(configure_one_file)
 set(SOURCE_FILES
   "__init__.py"
   "pass_groups/all.json"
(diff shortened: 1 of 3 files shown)


VitaNuo (approved):
Thank you!

emaxx-google:
thank you for this batch of reviews!

PR #250, opened 2025-07-25
pytype: Fix findings

emaxx-google:
Fix incorrect type signatures, add signatures where pytype would otherwise error out, and make overridden method signatures consistent.


VitaNuo (approved):
Thanks!

PR #251, opened 2025-07-25
pytype: Check in CI

emaxx-google:
Add a CI job to verify type signatures in Python code.


emaxx-google:
I had to fix one more typing issue that appeared in `master` after #250 went it.

PR #256, opened 2025-07-26
cli: Send file name log to intended output

emaxx-google:
Print the "--- \<testcase\> ---" line into the same output as all other precending&following logs - either stderr or the log file. It seems like it was an oversight to send only this log to stdout.

(We touched this code in #244, but the discrepancy was in the code before that change.)


VitaNuo (approved):
Thanks!

PR #257, opened 2025-07-28
test_manager: Decouple process_done_futures and job checks

emaxx-google:
A pure refactoring to make nested logic more readable and prepare it for adding new functionality to it.


VitaNuo (approved):
Thanks!

PR #258, opened 2025-07-28
interleaving: Track timeouts for each pass separately

emaxx-google:
Apply the limit "at most 20 timeouts before abandoning" per each
individual pass, as opposed to all passes in the interleaving group.

That'd avoid the problem that one frequently-timing-out pass aborts
the whole group of passes.


VitaNuo (approved):
Thanks!

PR #261, opened 2025-07-29
interleaving: Concurrently execute advance_on_success

emaxx-google:
Each time we commit to a reduction, schedule all passes' advance_on_success() methods to be executed in workers in parallel, as opposed to running them sequentially.

This significantly improves the performance of the "interleaving" mode, since we previously were spending 20-40% of wall clock time in those single-job advance_on_success() calls for heavy inputs.

Example measurements on a 64-core machine for the "gcc-94937" reproducer:
* before: reduced 54.4% after 5 min, 78.7% after 10 min.
* after: reduced 58.1% after 5 min, 85.2% after 10 min.

### DIFF
--- a/cvise/utils/testing.py
+++ b/cvise/utils/testing.py
@@ -888,7 +888,7 @@ class TestManager:
         ctx = self.pass_contexts[pass_id]
         assert ctx.can_init_now()
 
-        # Either initialize the pass from scract, or advance from the previous state.
+        # Either initialize the pass from scratch, or advance from the previous state.
         if ctx.state is None:
             env = InitEnvironment(
                 pass_new=ctx.pass_.new,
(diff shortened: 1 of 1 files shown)


cvise/utils/testing.py
```diff
@@ -830,8 +848,8 @@ def maybe_schedule_job(self, pool: pebble.ProcessPool) -> bool:
                 self.schedule_fold(pool, folding_state)
                 return True
         # 3. Attempting a transformation using the next heuristic in the round-robin fashion.
-        if any(ctx.state is not None for ctx in self.pass_contexts):
-            while self.pass_contexts[self.next_pass_id].state is None:
+        if any(ctx.stage == PassStage.ENUMERATING and ctx.state is not None for ctx in self.pass_contexts):
+            while self.pass_contexts[self.next_pass_id].state is None or self.pass_contexts[self.next_pass_id].stage != PassStage.ENUMERATING:
```
emaxx-google, line 852: Note: This is hardly readable, but the refactoring in #257 will make it much more concise.

PR #262, opened 2025-07-29
Override child processes' temp dir

### LINKED ISSUE #260: Temporary files not cleared sometimes
This is observed in the flakiness of the integration test added in #252, e.g.:

```
FAILED tests/test_cvise.py::test_simple_reduction_no_interleaving_config - AssertionError: assert [PosixPath('/tmp/pytest-of-root/pytest-0/test_simple_reduction_no_inter0/tmpdir/ccT9pAPp.s'), PosixPath('/tmp/pytest-of-root/pytest-0/test_simple_reduction_no_inter0/tmpdir/ccSDGNXT.s')] == []
  
  Left contains 2 more items, first extra item: PosixPath('/tmp/pytest-of-root/pytest-0/test_simple_reduction_no_inter0/tmpdir/ccT9pAPp.s')
  
  Full diff:
  - []
  + [
  +     PosixPath('/tmp/pytest-of-root/pytest-0/test_simple_reduction_no_inter0/tmpdir/ccT9pAPp.s'),
  +     PosixPath('/tmp/pytest-of-root/pytest-0/test_simple_reduction_no_inter0/tmpdir/ccSDGNXT.s'),
  + ]
```

This might be similar to the problem previously reported in #144.

emaxx-google:
When executing an interestingness test, make it use a custom temp dir, located inside the job's temporary folder, instead of the default location (like /tmp/ on a typical POSIX system). This allows C-Vise to clean up all temporary files it might've created without cluttering the default location.

Examples of problematic cases which this commit attempts to solve:

* The interestingness script launches GCC which creates a temporary `.s` file; meanwhile it's normally removed on GCC process exit, it doesn't happen when C-Vise kills cancelled jobs via SIGKILL.
* The interestingness script launches Clang without `-fno-crash-diagnostics`; once Clang crashes, it writes a the crash reproducer file to the temporary directory.

This fixes #260 (hopefully).

### DIFF
--- a/cvise/passes/abstract.py
+++ b/cvise/passes/abstract.py
@@ -122,7 +122,7 @@ class ProcessEventNotifier:
     def __init__(self, pid_queue):
         self.pid_queue = pid_queue
 
-    def run_process(self, cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=False):
+    def run_process(self, cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=False, env=None):
         if shell:
             assert isinstance(cmd, str)
         proc = subprocess.Popen(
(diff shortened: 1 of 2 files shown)


emaxx-google:
Thank you for the batch of the reviews.

PR #263, opened 2025-07-29
hints: Migrate clex::rm-toks pass

emaxx-google:
Port the "rm-toks-N" command from the "clex" pass - the one that removes N consecutive tokens based on a (simplistic) C grammar. The logic remains the same as before - the pass is just switched to using hints.

Here we also introduce a new parameter into the hint bundle - "grouping", so that we can disable the binary search logic for this pass "clex::rm-toks-N" (it'd be wasteful - there are other passes that are much better at recognizing big removable chunks of input).


emaxx-google:
Actually retracting this PR - this is so terribly inefficient now that I think about it.

Instead of running the parser 16 times and generating 16 hint bundles of form [s_i..s_{i+j}] for j=1..16 which are then tried one-by-one, we should just have a single bundle and have a logic that tries reasonable consecutive ranges.

emaxx-google:
> Actually retracting this PR - this is so terribly inefficient now that I think about it.
> 
> Instead of running the parser 16 times and generating 16 hint bundles of form [s_i..s_{i+j}] for j=1..16 which are then tried one-by-one, we should just have a single bundle and have a logic that tries reasonable consecutive ranges.

Reworked.

VitaNuo (approved):
Thanks!

PR #264, opened 2025-07-29
hints: More informative logs from ClangHints

emaxx-google:
Include the transformation name (like "remove-unused-function") into the logs - it's otherwise unclear how to distinguish between different ClangHints within an interleaving group.

Don't print warning logs about clang_delta failures as long as at least one of the --std= alternative succeeds. These warnings are misleading for the users, but it's normal that, e.g., a modern C++ program fails to compile in the --std=c++98 mode.

### DIFF
--- a/cvise/passes/clanghints.py
+++ b/cvise/passes/clanghints.py
@@ -27,6 +27,10 @@ class ClangState(HintState):
         return wrapped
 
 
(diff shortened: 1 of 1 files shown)


VitaNuo (approved):
Thank you!

PR #267, opened 2025-07-31
tree-sitter: Import 0.20.8

emaxx-google:
Add a new third-party dependency: Tree-sitter. We're going to create new reduction heuristics based on Tree-sitter parsers.

Note 1: We import the code instead of relying on system-wide packages, since their availability varies across platforms and because of possible ABI incompatibilities. We don't use git submodules because a monorepo is easier to maintain and has better tooling support.

Note 2: We use the outdated version 0.20.8 for easier compatibility story with downstream projects; in the future it can be rolled to a newer version, but it's not critical for the initial functionality.


emaxx-google:
@VitaNuo Asking only the high-level review - all files in the PR are simply the unpacked https://github.com/tree-sitter/tree-sitter/archive/refs/tags/v0.20.8.tar.gz .

PR #270, opened 2025-07-31
tree-sitter-cpp: Build rule

emaxx-google:
Trigger building the Tree-sitter C++ grammar.

We simply reuse the CMake source file from the upstream. The only tweak is disabling dynamic library creation, which the upstream CMake sets to "on" by default.

On the C-Vise side, we stop applying "-pedantic" unilaterally, since it doesn't work for Tree-sitter subdirectories.


VitaNuo (approved):
Thank you!

PR #274, opened 2025-08-04
interleaving: Reinit finished passes once in a while

emaxx-google:
After a pass finished enumerating all of its state, reinitialize it again
if we made reductions via other passes (in the interleaving mode).

This significantly improves the reduction pace in its mid-term stage,
when some passes already went through all of the candidates once
but there's still a couple of passes that continue making successful
reductions. For example, `LinesPass` is likely to finish before `rm-tok`
(as there's more tokens that curly brace pairs), however removing
some token can make some curly-brace-surrounded block
removable too - this is where the reinitialization of `LinesPass`
would help discovering this sooner, within the same batch of jobs
in run_passes().

We just need to make sure we don't do reinits too often, since for
some passes they're mostly useless (think of `CommentsPass` after
all comments removed), and even for other passes only occasionally
useful. Our choice is - don't dedicate more than 1% of jobs to this.

Empirical measurements on our test dataset: C-Vise running time
improvement around 20% (varying between 0% and 75%).

### DIFF
--- a/cvise/utils/testing.py
+++ b/cvise/utils/testing.py
@@ -308,6 +308,9 @@ class TestManager:
     BUG_DIR_PREFIX = 'cvise_bug_'
     EXTRA_DIR_PREFIX = 'cvise_extra_'
     EVENT_LOOP_TIMEOUT = 1  # seconds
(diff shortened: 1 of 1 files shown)


emaxx-google:
thank you for the reviews!

PR #276, opened 2025-08-05
clang_delta: Skip non-dep typedefs in replace-dependent-typedef

emaxx-google:
Only consider typedefs that depend on other typedefs in replace-dependent-typedef.

This prevents the pass from reporting a success when no actual change has been made, like in the case when a typedef belongs to and refers to a template struct (where multiple possible string representations may confuse the checks).


VitaNuo (approved):
Thanks!

PR #278, opened 2025-08-07
tree-sitter: Add ReplaceFunctionDefWithDecl pass

emaxx-google:
Implement a Tree-sitter parser based pass that replaces function definitions with declarations - either replacing function bodies with a semicolon or, in case of out-of-line members, deleting the whole definition. Constructor initializer lists are another special case.

Note: meanwhile this new pass effectively duplicates the existing clang_delta "replace-function-def-with-decl" heuristic, the new pass is faster, doesn't require the "-std=" guesswork and, most importantly, can be easily reused for multi-file reproducers (unlike Clang, Tree-sitter doesn't require #include search paths or any other compilation parameters).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,6 +225,7 @@ add_subdirectory(cvise)
 add_subdirectory(delta)
 add_subdirectory(tree-sitter)
 add_subdirectory(tree-sitter-cpp)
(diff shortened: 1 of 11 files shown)


VitaNuo (approved):
Thank you!

emaxx-google:
thank you for the reviews

PR #285, opened 2025-08-12
tree-sitter, tree-sitter-cpp: Don't install into system location

### LINKED ISSUE #284: Don't install bundled tree-sitter-cpp
cvise master now builds and installs a bundled copy of tree-sitter-cpp, which collides with a system copy if it is packaged:
```
 * Detected file collision(s):
 *
 *      /usr/include/tree_sitter/tree-sitter-cpp.h
 *      /usr/lib64/libtree-sitter-cpp.so
 *
 * Searching all installed packages for file collisions...
 *
 * Press Ctrl-C to Stop
 *
 * dev-libs/tree-sitter-cpp-0.23.1:0::gentoo
 *      /usr/include/tree_sitter/tree-sitter-cpp.h
 *      /usr/lib64/libtree-sitter-cpp.so
 *
 * Package 'dev-util/cvise-9999' NOT merged due to file collisions. If
 * necessary, refer to your elog messages for the whole content of the
 * above message.
 *
 * The following package has failed to build, install, or execute postinst:
 *
 *  (dev-util/cvise-9999:0/0::gentoo, ebuild scheduled for merge), Log file:
 *   '/var/log/portage/build/dev-util/cvise-9999:20250812-055736.log'
 *
```

cvise's build should:
1) search using `pkg_check_modules` for a system copy
1) if that doesn't exist, build and statically link the bundled copy, and do not install any files from it

emaxx-google:
Don't execute by default all CMake rules in tree-sitter and tree-sitter-cpp - instead, only those rules will be executed that are depended by something else (currently only treesitter_delta).

This fixes the problem that installing C-Vise would result in installing tree-sitter-cpp public .h/.so files into system-wide locations. This fixes #284.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,8 +223,8 @@ add_subdirectory(clang_delta)
 add_subdirectory(clex)
 add_subdirectory(cvise)
 add_subdirectory(delta)
-add_subdirectory(tree-sitter)
-add_subdirectory(tree-sitter-cpp)
+add_subdirectory(tree-sitter EXCLUDE_FROM_ALL)
+add_subdirectory(tree-sitter-cpp EXCLUDE_FROM_ALL)
 add_subdirectory(treesitter_delta)
 
 # Copy top-level cvise script


VitaNuo (approved):
Thanks!

PR #286, opened 2025-08-14
Make State objects immutable and hashable

emaxx-google:
Change a few "state" classes to let them be used for comparison and for hashing containers.

This refactoring should have no functional effect, but it's a prerequisite for a follow-up change where state comparisons will be necessary.


emaxx-google:
thanks for reviews!

PR #294, opened 2025-08-18
multifile: Add support to Lines pass

emaxx-google:
Migrate the Lines pass to support the multi-file (i.e., directory instead of a single file) inputs.


VitaNuo (approved):
Thanks!

PR #295, opened 2025-08-19
logging: Collect logs in main process, filter out canceled jobs

emaxx-google:
Send logs from worker processes over an IPC pipe instead of printing them directly to stderr. Discard logs coming from canceled jobs.

This is a more robust approach than the previous silent discarding of all OSError exceptions. Also it fixes the problem with --log-file due to workers unable to write to this file concurrently. Additionally, it allows us to implement more aggressive worker reuse and cancellation in follow-up PRs.


VitaNuo (approved):
Thanks!

PR #296, opened 2025-08-19
Fix flakiness in test_cvise.py - trailing line break

emaxx-google:
The tests were occasionally failing because the result may or may not terminate with a newline character.


VitaNuo (approved):
Thanks!

PR #297, opened 2025-08-19
multiprocessing: Don't create pool many times

emaxx-google:
Have a single process pool instead of recreating it every time pass(es)
are to be executed. We only need to make sure to cancel running jobs.

This brings a significant speedup especially for small cases or for late
stages of reductions in general, where it takes just a fraction of a
second to discover a reduction. Now we don't have to pay the price for
initializing the Pebble pool - hundreds of milliseconds - repeatedly.

While empirical measurements show 1.5x-2x speedup, unfortunately
most of it will be "eaten" by the switch to "forkserver" multiprocessing
mode - the latter seems unavoidable to fix occasional deadlocks.


VitaNuo (approved):
Thanks!

PR #299, opened 2025-08-21
test_cvise: Use temp dir instead of in-tree files

emaxx-google:
Slightly improve test_cvise.py to not create temporary files in its own location, and use temporary directories instead.

This makes the tests more reliable and compatible with build systems like Bazel that don't allow modifying input files. Also it simplifies the implementation of multi-file (directory) reduction tests, where the cleanup would be more annoying to write.


VitaNuo (approved):
Thanks!

PR #301, opened 2025-08-21
multifile: Introduce support into CLI, test_manager, passes

emaxx-google:
Introduce the most basic implementation of multi-file reductions (minimizing whole directories instead of separate files).

Only two heuristics are supported as of this commit: Comments and Lines. Caching and backup functionality doesn't work at the moment.


VitaNuo (approved):
Thanks!

PR #302, opened 2025-08-21
Gracefully clean up on SIGTERM

emaxx-google:
Monitor and handle SIGTERM in addition to SIGINT.

First, this prevents C-Vise from leaving temporary files when killed by SIGTERM. Second, we want pebble worker processes to do proper cleanup when jobs are canceled.


VitaNuo (approved):
Thank you!

PR #304, opened 2025-08-22
test_manager: Replace files atomically when committing

emaxx-google:
Instead of directly writing to the test case, write first to a temporary file and then rename it. This is atomic on popular file systems.

Also refactor the code to have a dedicated fileutil helper, and cover it with unit tests. It'll make the upcoming support of directory inputs in #301 slightly cleaner.


VitaNuo (approved):
Thanks!

emaxx-google:
After playing more with this, I've discovered the previous solution with `CloseableTemporaryFile` wasn't fully robust. It took me some time to implement a test to relatively reliably catch concurrency issues, and it was fun to see how it breaks many of my attempts... The new version seems to be actually more robust.

PR #305, opened 2025-08-22
multiprocessing: Use forkserver mode

emaxx-google:
This fixes potential deadlocks caused by forking a multi-threaded program,
and follows generic Python guideline. Also it reduces the RAM pressure since
workers don't get copies of the main C-Vise process data structures, like the
pass cache (despite copy-on-write not literally copying those pages, programs
like OOM monitors might treat this as real usage).

The downside is that it causes a 25-50% slowdown compared to the "fork" mode,
because forkserver workers are slower to start and interact with. Part of this
seems to be caused by https://github.com/python/cpython/issues/137996,
which will hopefully fixed in new Python versions. The rest of slowdown we'll
mitigate by further optimizations in C-Vise in follow-up commits.

### DIFF
--- a/cvise-cli.py
+++ b/cvise-cli.py
@@ -6,6 +6,7 @@ import datetime
 import importlib.util
 from itertools import chain
 import logging
+import multiprocessing
 import os
 import os.path
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


VitaNuo (approved):
Thanks!

PR #306, opened 2025-08-24
tree-sitter: Always link statically

emaxx-google:
Ignore the externally specified BUILD_SHARED_LIBS value when building tree-sitter and tree-sitter-cpp, and link them statically always. This is another fix for #284.


VitaNuo (approved):
Thank you!

PR #307, opened 2025-08-25
test_manager: Quickly exit worker on signal outside jobs

emaxx-google:
Stress-testing showed that raising KeyboardInterrupt/SystemExit exceptions at unpredictable time points, especially inside multiprocessing internals, can put the worker into a dysfunctional state, which may lead to deadlocks.

This commit employs a different strategy: exception-based signal propagation in worker processes is only used when executing a job; this lets the job gracefully release resources, like kill spawned subprocesses. Once the job completes or when a signal is received without any job, the process is terminated immediately.

### DIFF
--- a/cvise/cvise.py
+++ b/cvise/cvise.py
@@ -94,7 +94,7 @@ class CVise:
     }
 
     def __init__(self, test_manager, skip_interestingness_test_check):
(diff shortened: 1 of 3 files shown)


VitaNuo (approved):
Thanks!

PR #308, opened 2025-08-26
test_cvise: Use a shorter TMPDIR

emaxx-google:
For the TMPDIR value that we pass when executing C-Vise in integration tests, create a fresh temp dir in a regular way (e.g., under /tmp/ on a typical *nix) instead of stacking one inside the pytest temp path.

This shorter TMPDIR value reduces the possibility of hitting the "path too long" errors, especially when using pytest-xdist, which makes pytest tmp_path longer.

### DIFF
--- a/tests/test_cvise.py
+++ b/tests/test_cvise.py
@@ -5,43 +5,47 @@ import shutil
 import signal
 import stat
 import subprocess
+import tempfile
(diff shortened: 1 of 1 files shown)


VitaNuo (approved):
Thanks!

PR #311, opened 2025-08-26
subprocesses: Kill via PID queue more often

emaxx-google:
Kill "leftover" child processes - as reported via started, but not finished, events in the PID queue - every time we stop worker jobs.

Before this commit, this only happened after all jobs for a particular Pass have finished, which could take a very long time - and orphan child processes could've continued working, eating machine resources and potentially leading to OOMs.


VitaNuo (approved):
Thanks!

emaxx-google:
I think I'll need to rework this PR, since after #297 `Don't create pool many times`, we don't wait for worker termination when proceeding with a reduction, hence the SIGKILLS coming from the C-Vise main process would "race" with SIGTERMs sent by the shutting-down Pebble workers. We generally want SIGTERMs as the preferred, graceful way of shutting down child process hierarchies, so this PR should somehow defer sending SIGKILLs until we know the corresponding worker is terminated.

This isn't trivial to implement, since Pebble doesn't expose details of worker processes, but I think we can do it by creating a thin wrapper around the `multiprocessing` context, with additional hooks for our logic.

PR #314, opened 2025-08-27
logging: Fix breakage when worker logs an exception

emaxx-google:
This fixes the "TypeError: cannot pickle 'traceback' object" error. It was introduced together with the "mplogging" logic in #295.

We should transmit an already-formatted log record over IPC, without unpickleable fields.


VitaNuo (approved):
Thanks!

PR #315, opened 2025-08-28
multifile: Add support to Tree-sitter pass

emaxx-google:
Migrate all Tree-sitter based heuristics (replace-function-def-with-decl, erase-namespace, remove-function) to support multi-file inputs, i.e. test cases that are directories.

Note: "remove-function" is the first truly cross-file heuristic, as it's attempting to remove all function defs/decls sharing the same name from all files.


VitaNuo (approved):
Thank you!

PR #317, opened 2025-08-28
test_cvise: Use specific worker count

emaxx-google:
Hardcode a specific '-n' value in the test, so that it doesn't deviate significantly depending on how beefy the machine is. Also graceful worker termination would take more time than the hardcoded timeout for a very big number of workers.


VitaNuo (approved):
Thank you!

PR #318, opened 2025-08-28
tree-sitter: Import 0.25.8

emaxx-google:
Update the version of the Tree-sitter library to the currently latest release.

This brings some bugfixes, also hopefully it'll silence security warnings in the C-Vise Github repo.


emaxx-google:
@VitaNuo The library is imported as-is, without any modifications (but our custom CMakeLists.txt remains).

VitaNuo:
Ok, thanks. This makes sense for the upstream. How about the internal version? We'll have to link against the only version we have available, and we don't even have proper control over which version that is. Is that likely to cause problems? Do we need a more robust solution for that?

emaxx-google:
> Ok, thanks. This makes sense for the upstream. How about the internal version? We'll have to link against the only version we have available, and we don't even have proper control over which version that is. Is that likely to cause problems? Do we need a more robust solution for that?

Yes, there's a small potential concern for the Google internal monorepo (or any other monorepo with the "one version rule" like the https://opensource.google/documentation/reference/thirdparty/oneversion). Note that ABI is not a concern for a monorepo, since each library exists in one version that everything else links against. So we'll only be concerned about grammar backwards-incompatible changes; but they should be relatively rare and hopefully they should be caught by C-Vise tests - at which point the person who updates the grammar should notice it and make necessary changes in C-Vise as well (again a benefit of a monorepo - being able to do atomic changes like this).

So TLDR is that I think we shouldn't worry too much about this from the downstream perspective. But let me know what you think!

VitaNuo:
Ok, thanks for the explanation!

VitaNuo (approved):

PR #319, opened 2025-08-29
subprocesses: Fix busyloop after terminating child

emaxx-google:
Actually exit the wait loop when the subprocess wait() succeeds, instead of busy-looping until the timeout passes.

This manifested itself as some test timeouts.


VitaNuo (approved):
Thanks!

PR #323, opened 2025-08-30
CI: Build with larger parallelism than nproc

emaxx-google:


emaxx-google:
This doesn't help - looks like the CI worker resource utilization is already good.

PR #325, opened 2025-09-02
multiprocessing: Track worker processes and their children

emaxx-google:
Implement hooks that allow us knowing Pebble worker PIDs, and be notified of their startup&termination. Track down subprocess PIDs started by each worker.

This is preparation work for quickly killing children of canceled/timed-out Pebble jobs. Also it paves the way to reliably killing processes via process groups (on *nix), measuring and limiting RAM usage by worker's job, etc.


emaxx-google:
thanks for this batch of reviews! also added tests here.

PR #327, opened 2025-09-02
multiprocessing: Stop all jobs before doing pool.__exit__

emaxx-google:
In order to exit quickly on Ctrl-C or other signals, we should switch the Pebble process pool into the stopped mode, which cancels all jobs and starts shutting down all workers.


VitaNuo (approved):
Thanks!

PR #329, opened 2025-09-04
multiprocessing: Workaround for hung Pebble jobs

emaxx-google:
Pebble seems to have a race condition, resulting in jobs occasionally "hanging": the future never gets resolved with a result nor with a timeout. This results in hanging the whole C-Vise process, since it waits for a completion of a job infinitely long.

As a workaround, we attempt to detect hung jobs by looking at those whose timing violate the expected duration severely. We cancel such tasks, which potentially means we miss some reduction opportunities - but this should be rare, and the other loop over heuristics should likely rediscover the reduction again.

We also introduce a (more generous) timeout for pass init jobs, since without it we wouldn't be able to detect hangs of such jobs.


VitaNuo (approved):
Thanks!

PR #330, opened 2025-09-04
Add timeouts for pass initialization jobs

emaxx-google:
Impose a timeout for pass initializations, so that a hung pass never hangs the whole C-Vise process. Additionally, this change allows us to gracefully handle the bug in Pebble when a scheduled task may occasionally be lost.

We set the init timeout to be a generous 10x of the regular --timeout value, as a safe default without the user having to guesstimate the parameter themselves.


VitaNuo (approved):
Thank you!

PR #331, opened 2025-09-05
multifile: Add Clang .modulemap heuristic

emaxx-google:
Implement a new heuristic ClangModuleMap that's capable of removing chunks of a module map (aka .cppmap) file in a structured way.

Note: this commit uses a simple handwritten parser that only supports a practically common subset of the module map grammar; in the future it should probably be replaced with a proper robust parser.


VitaNuo (approved):
Thank you!

PR #332, opened 2025-09-05
Fix flaky test_test_manager: wait for mgr shutdown

emaxx-google:
Before asserting the state at the end of the test, make sure to shut down the manager to let multi-threaded logic complete.


VitaNuo (approved):
Thank you!

PR #333, opened 2025-09-09
subprocesses: Make PID queue robuster to signals

emaxx-google:
Don't allow signals to interrupt run_process() in a way that neither FINISHED nor ORPHANED are returned.

Refactor the sigmonitor code so that various modes (os._exit, raise, remember-but-raise-on-demand-only) are easier to control and to reduce the runtime overhead if we add it into hotpaths.

### DIFF
--- a/cvise/cvise.py
+++ b/cvise/cvise.py
@@ -96,7 +96,7 @@ class CVise:
     }
 
     def __init__(self, test_manager, skip_interestingness_test_check):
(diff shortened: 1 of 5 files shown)


emaxx-google:
This also contains some non-pebble-specific bits of #329 extracted into a separate PR.

emaxx-google:
thanks for the reviews!

PR #334, opened 2025-09-09
hints: Passes declare output hint types

emaxx-google:
Each pass now explicitly declares types of the hints it produces.

This will be used to implement pass dependencies - some passes consuming the information produced by others.


VitaNuo (approved):
Thank you!

PR #338, opened 2025-09-11
hint: Optimize byte ops in i/o and patch applying

emaxx-google:
Use bytearray to avoid repeatedly copying buffers when concatenating chunks for every single patch. Store vocabulary as bytes to avoid doing UTF-8 encoding on every patch with replacement ("v"). Remove ifs from the hint loading inner loop.

Measurements show 5-10% speedup on big inputs (>10MB).


emaxx-google:
After a couple of more small optimizations, this is ready for landing I think.

PR #339, opened 2025-09-14
hints: Use msgspec structs as optimization 

emaxx-google:
Introduce proper classes for the "hint" and "patch" concepts, based
on msgspec's structs - instead of using Python dictionaries.

This is an optimization - this makes the bare reading/writing of hints
10-20% faster, and makes C-Vise around 3% faster (also the first few
reduction steps finish more noticeably faster). Also proper classes
make the code more readable and prevent typos in Python code of passes.


VitaNuo (approved):
Thanks!

emaxx-google:
Further polished it to use `gc=False` mode, use a more optimized hint grouping, reuse decoders.

PR #353, opened 2025-09-23
multifile: Forbid interestingness script inside test case

emaxx-google:
Check at the very start whether the user-specified interestingness script lies inside any of the test cases - this is unsupported and likely to bring troubles (as a randomly-modified shell script won't perform any useful check, and may even be dangerous to run).


VitaNuo (approved):
Thanks!

emaxx-google:
thanks for taking a look!

PR #357, opened 2025-09-24
test_manager: Giveup is per pass, not for whole category

emaxx-google:
Apply the "giveup" limit - assume "pass got stuck" after 50000 unsuccessful jobs in a row - to an individual pass rather than to the whole pass group.

This allows the "interleaving" mode to continue making progress using some passes even if others got stuck and shouldn't be used anymore.


VitaNuo (approved):
Thanks!

PR #359, opened 2025-09-24
multifile: Introduce RmUnusedFiles pass

emaxx-google:
This pass consumes `@fileref` hints from other passes (like ClangModuleMap, ClangIncludeGraph, etc.) and attempts deleting files not referred by anyone.

We also extend the hints schema to have a new field "o" ("operation"), allowing us to denote the "rm" (file deletion) operation - previously hints could only express file content insertions/removals.


VitaNuo (approved):
Thanks!

PR #360, opened 2025-09-25
multifile: Assume all Makefiles to be used

emaxx-google:
Suppress removal of Makefiles in the RmUnusedFilesPass, because in a typical use case the makefile would be used by the interestingness test to compile the source.

An ideal solution would be to detect whether a makefile is actually used during the interestingness test's invocation, but currently C-Vise cannot make any insights into the interestingness tests. In the future an strace-like solution might be considered.


emaxx-google:
thanks for taking a look!

PR #363, opened 2025-09-25
multifile: MakefilePass: Fix parsing of args with quotes

emaxx-google:
Fix an assertion failure when parsing a command line with quotes in the middle of arguments, like:

`  -Dfoo=\"bar\"`

The problem was that when using the parser from "shlex", we received tokens with quotes removed/unescaped, and we couldn't identify their locations in the original command (which is necessary for passes like the command line arg removal).

To fix the problem, we implement a custom handwritten parser of shell-like command lines.


VitaNuo (approved):
Thank you!

PR #364, opened 2025-09-26
multifile: Fix wrong paths in Lines; more robust path handling

emaxx-google:
Fix the LinesPass producing malformed hints for directory (multi-file) test cases, because the Python side didn't pass relative paths.

Fix file opening error handling in LinesPass and ClexHintsPass to not accidentally fall back to reading stdin (previously we left yyin as NULL and proceeded to running Flex, which then by default fell back to reading from stdin).

Make LinesPass, ClexHintsPass and TreesitterPass robust to file names with newlines by using null characters to delimit file names (this approach was already used for ClangIncludeGraphPass).

In tests, assert additionally that each hint patch doesn't exceed file boundaries.

### DIFF
--- a/clex/clex.l
+++ b/clex/clex.l
@@ -165,6 +165,7 @@ int count = 0;
 // index of byte where the current yytext ends in the input.
 int tok_end_pos = 0;
 
-void reset_to_initial_state(void) {
+void restart_with_new_file(void) {
     BEGIN(INITIAL);
+    YY_NEW_FILE;
 }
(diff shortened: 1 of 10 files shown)


VitaNuo (approved):
Thank you!

MehdiChinoune:
getdelim is not a standrad C function and It's available on Windows (both MSVC and MinGW)

emaxx-google:
> getdelim is not a standrad C function and It's available on Windows (both MSVC and MinGW)

should be fixed (#481); thanks for reporting!

PR #365, opened 2025-09-26
multifile: Fix Clex rm-toks crash on unended comments

emaxx-google:
Fix the ClexHintsPass returning no hints when a directory (multi-file) test case has files with unfinished C-style comments.

Implementation-wise, there were two issues: first, Flex doesn't report EOF from input() when it was the last file and yywrap() куегктув тщтяукщ; repeatedly calling input() was causing a segfault. Second, we should stop the loop that "eats" comments when we cross input file boundary.

### DIFF
--- a/clex/clex.l
+++ b/clex/clex.l
@@ -22,6 +22,7 @@ IS			(u|U|l|L)*
 #endif
 
 #include <defs.h>
+#include <limits.h>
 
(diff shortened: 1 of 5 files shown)


emaxx-google:
I'll add a regression test for this a bit later.

PR #368, opened 2025-09-29
hints: Use tuples, make frozen and comparable

emaxx-google:
Change the Hint/Patch structs to be frozen and ordered, and consequently store patches in tuples instead of lists.

This should reduce RAM usage for passes that generate lots of hints. Additionally, it allows us to sort and remove duplicate hints easily.


VitaNuo (approved):
Thanks!

PR #369, opened 2025-09-29
ClangIncludeGraph: Parallelize the pass initialization

emaxx-google:
As running clang_include_graph sequentially may take a lot of time (several minutes) on big inputs, we split this work into K workers that are executed concurrently - this way we reduce the wall clock time it takes for the pass to finish initialization. Currently the hardcoded value of K is 10.

Implementation-wise, to avoid introducing new complexity if we were to make new()/advance_on_success() concurrent, we instead reuse the previously implemented mechanism of pass dependencies: we spawn K "subordinate" passes that do the actual work, and then the main pass just merges the hints produced by them all. This is a simply form of map-reduce, but built upon primitives we previously added to C-Vise.


VitaNuo (approved):
Thanks!

PR #373, opened 2025-09-30
Support for LLVM 22

emaxx-google:


emaxx-google:
After making the code compile, the first results of the tests: `12 failed, 176 passed`.

emaxx-google:
Down to 6 failures.

(FWIW most of troubles are caused by the gigantic change in Clang https://github.com/llvm/llvm-project/pull/147835.)

PR #375, opened 2025-10-01
multifile: MakefilePass removes targets

emaxx-google:
Implement a new heuristic that removes a makefile target together with all references to it (in prerequisites and command lines).

Also improve the "remove-arguments-across-all-commands" heuristic in the same pass to not leave extraneous spaces around the deleted arguments.


VitaNuo (approved):
Thanks!

PR #377, opened 2025-10-01
interleaving: Skip states that are subsets of successes

emaxx-google:
Don't spend time running an interestingness test for a pass state that is known to be a part of another state for which the interestingness test already returned success before.

For example, if CommentsPass's first job that removes all comments passes the interestingness test, we don't need to run the jobs trying the first or the second half of the comments (or any other job at all, in this example).


VitaNuo (approved):
Thank you!

PR #378, opened 2025-10-01
hints: Sort hints in all passes

emaxx-google:
Make sure hints are sorted lexicographically: by file, type, 'l'/'r' indices within the same file, etc.

This makes sure passes enumerate reduction possibilities in a stable and well-defined order.


VitaNuo (approved):
Thanks!

PR #380, opened 2025-10-02
multifile: Print file&dir count in logs

emaxx-google:
Logs for directory (multi-file) test cases will look like:

`00:30:26 INFO (58.7%, 9998801 bytes, 149447 lines, 2481 files in 675 dirs, ...)`


emaxx-google:
thanks for taking a look!

PR #381, opened 2025-10-02
cli: Stats sort passes by duration too

emaxx-google:
When multiple passes have the same improvement, sort them by increasing of total duration. This makes the stats table a bit more intuitive.


VitaNuo (approved):
Thanks!

PR #382, opened 2025-10-02
tests: Add another possible output into test_cvise

emaxx-google:
The semi-integration test of reducing blocksort-part.c can sometimes end up with yet another result: "#undef nextHi". This was causing occasional flakiness of the test.


VitaNuo (approved):
Thanks!

PR #383, opened 2025-10-02
hints, test_manager: Sanitize and randomize file names

emaxx-google:
Only allow regular characters (a-zA-Z0-9_.-) in file names that C-Vise creates for internal needs.

First, this fixes an annoying problem that the interestingness script could be executed in a folder with spaces in the names - and it's pretty hard for the user of C-Vise to discover if their Shell script doesn't support this. Second, it makes sure we don't use hint types - which can be arbitrary strings - to form malformed file names.

As a bonus, this commit simplifies implementing a better folding mechanism in the future, with reusing successful discoveries from a previous batch of jobs that didn't make it into the best folding attempt.

### DIFF
--- a/cvise/passes/abstract.py
+++ b/cvise/passes/abstract.py
@@ -206,6 +206,7 @@ class AbstractPass:
         self,
         test_case: Path,
         state,
+        new_tmp_dir: Path,
         succeeded_state,
         job_timeout,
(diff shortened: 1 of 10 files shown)


VitaNuo (approved):
Thank you!

PR #384, opened 2025-10-02
hints: Implement hint subtraction

emaxx-google:
This allows us to recalculate old hints after a reduction is applied, without having to invoke the pass initialization.

This will be used in follow-ups to implement a better folding mechanism, carrying over those successful transformations that didn't make it into the best folding attempt. Additionally, this could be used to implement a cheap version of advance_on_success() that's based on pure position computations (e.g., without the need to rebuild the Clang AST).


VitaNuo (approved):
Thanks!

PR #386, opened 2025-10-07
hints: Reland "Sort hints in all passes"

emaxx-google:
Compared to the original patch (#378), we now properly support comparing Nones with non-Nones in fields. Original patch description:

Make sure hints are sorted lexicographically: by file, type, 'l'/'r' indices within the same file, etc.

This makes sure passes enumerate reduction possibilities in a stable and well-defined order.


VitaNuo (approved):
Thank you!

PR #387, opened 2025-10-07
Clang passes: Pass params directly to constructor

emaxx-google:
This is a small refactoring to explicitly pass
ClangPass/ClangHintsPass/ClangBinarySearchPass parameters directly as constructor arguments, instead of the caller setting them as fields afterwards.

This makes the contract more explicit, and also prepares us for introducing more pass-specific construction parameters in follow-up commits.

### DIFF
--- a/cvise/cvise.py
+++ b/cvise/cvise.py
@@ -155,7 +155,8 @@ class CVise:
 
             pass_group[category.name] = []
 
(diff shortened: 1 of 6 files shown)


VitaNuo (approved):
Thank you!

PR #388, opened 2025-10-08
makefile,cppmap: Configure file patterns in config; skip other passes

emaxx-google:
Introduce a new property "claim_files" into the JSON config, specifying file glob patterns that should only be processed by particular passes; other passes will automatically skip these files.

This removes hardcoded file patterns from C-Vise Python code (except for makefileparser's logic for other passes that will be addressed later). Also this prevents "naive"/dangerous modifications by passes that don't understand the file structure.

### DIFF
--- a/cvise/tests/test_test_manager.py
+++ b/cvise/tests/test_test_manager.py
@@ -9,16 +9,17 @@ import pytest
 import re
 import sys
 import time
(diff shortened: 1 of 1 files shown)


VitaNuo (approved):
Thank you!

PR #507, opened 2026-06-15
Fix test_dir_linker_duplicate_var_error on non-english locales

giulianobelinassi:
If the host machine is running another locale language other than english, the test `test_dir_linker_duplicate_var_error` fails because the searched string is not found.  Fix this by enforcing that make is called with `LC_ALL=C`.


emaxx-google:
thanks!

emaxx-google (approved):

PR #509, opened 2026-06-29
subprocesses: Exit job on orphan i/o streams

### LINKED ISSUE #508: Test `test_dir_linker_duplicate_var_error` hangs on single core VMs
The test ` test_dir_linker_duplicate_var_error` gets stuck when running on a single core virtual machine:

```
260418.64ff7de2/build> pytest tests/test_cvise.py::test_dir_linker_duplicate_var_error -n 1
==================================== test session starts =====================================
platform linux -- Python 3.13.13, pytest-9.0.3, pluggy-1.6.0 -- /usr/bin/python3.13
cachedir: .pytest_cache
rootdir: /home/abuild/rpmbuild/BUILD/cvise-2.12.0+git.20260418.64ff7de2-build/cvise-2.12.0+git
.20260418.64ff7de2/build
configfile: setup.cfg
plugins: mock-3.15.1, subprocess-1.6.0, xdist-3.8.0
1 worker [1 item]      
scheduling tests via LoadScheduling

tests/test_cvise.py::test_dir_linker_duplicate_var_error 
****
```
and it keeps stuck there infinitely. Running with `-n 2` makes the test pass.  All my attempts to reproduce it on the host machine without the QEMU virtual machine were unsuccessful: the test always passed.

emaxx-google:
Prevent hanging a job until reaching the timeout in case the interestingness test's process exits, however its stdout/stderr streams remain open (e.g., due to child processes).

Fixes #508

### DIFF
--- a/cvise/utils/process.py
+++ b/cvise/utils/process.py
@@ -358,6 +358,13 @@ class ProcessEventNotifier:
             except subprocess.TimeoutExpired:
                 if step_timeout == 0:
                     raise  # we reached the original timeout, so bail out
+
+                # If the child process has already exited but orphaned descendants keep the pipes open, communicate()
+                # will wait for EOF forever. In this case, we close the pipes and bail out.
+                if proc.poll() is not None:
+                    _close_pipes(proc)
+                    return proc.communicate(input=b'')  # type: ignore[arg-type]
+
                 input = b''  # the input has been written in the first communicate() call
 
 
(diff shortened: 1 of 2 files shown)


emaxx-google:
There's actually an interesting tradeoff here, because if we let the C-Vise job complete despite the background processes that linger around, we're potentially exceeding the `-n` limit specified by the user. This might be very undesired by the end users since it might cause resource exhaustion on the machine...

Ideally if we go that route, we should have logic in place that automatically kills descendant processes on a job completion. Which would probably require a per-job mechanism for tracking descendant processes (like `start_new_session=True`).

